PR #64, opened 2019-10-25
Removing mentions about Rector CI

JanMikes:
Leave this great idea for later


TomasVotruba:
Just like Packagist, this might come in the future

image

PR #71, opened 2019-11-22
Fix typos on homepage

ikvasnica:
Just fixes of little typos ;)


TomasVotruba:
We have some troubles with automated deploy, that's why it got stuck.

For now I've deplyoed it manually, so fixes are on :+1: 

image

PR #80, opened 2020-01-16
Require rector/rector dependency in non-dev

JanMikes:


JanMikes:
Just fyi, it was required because we are using `SetProvider` in normal application and it was failing on missing rector/rector package in production :-)

PR #91, opened 2020-01-19
More space for demo, add syntax highlight

TomasVotruba:
## Before

image

## After

image


src/Controller/DemoController.php
```diff
@@ -36,55 +36,68 @@ final class DemoController extends AbstractController
...
+            $formData->setContent($rectorRun->getContent());
+            $formData->setConfig($rectorRun->getConfig());
+        } else {
+            // default values
+            $demoFileContent = FileSystem::read(__DIR__ . '/../../data/DemoFile.php');
+            $demoConfig = FileSystem::read(__DIR__ . '/../../data/demo-config.yaml');
+
+            $formData->setContent($demoFileContent);
+            $formData->setConfig($demoConfig);
+        }
```
TomasVotruba: I wonder if those 2 parameters could be required arguments in `DemoFormData`.
That would remove setters and make contract mor?e clear. Could you try?

src/Controller/DemoController.php
```diff
@@ -36,55 +36,68 @@ final class DemoController extends AbstractController
...
     {
-        /** @var RectorRunFormData $formData */
-        $formData = $form->getData();
-        $config = $formData->getConfig();
+        /** @var DemoFormData $rectorRunFormData */
+        $rectorRunFormData = $form->getData();
+        $config = $rectorRunFormData->getConfig();
 
-        $rectorRun = new RectorRun(Uuid::uuid4(), new DateTimeImmutable(), $config, $formData->getContent());
+        $rectorRun = $this->createRectorRun($config, $rectorRunFormData);
```
TomasVotruba: Basically just renamed `RectorRun*` to `Demo*`, for consistency with Form, Controller and template

config/services.yaml
```diff
@@ -13,10 +13,16 @@ services:
         resource: '../src/'
         exclude:
             - '../src/GetRectorKernel.php'
-            - '../src/Entity'
-            - '*FormData.php'
-            - '*Exception.php'
+            - '../src/Entity/*'
+            - '../src/ValueObject/*'
+            - '../src/Exception/*'
```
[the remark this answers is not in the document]
JanMikes: Ok, i see, i am used to it through DDD, where you couple code by domain and not by it's type :-)

src/Controller/DemoController.php
```diff
@@ -36,55 +36,68 @@ final class DemoController extends AbstractController
...
 
         if ($form->isSubmitted() && $form->isValid()) {
             return $this->processFormAndReturnRoute($form);
         }
 
         return $this->render('homepage/demo.twig', [
-            'form' => $form->createView(),
-            'rector_run' => $rectorRun,
+            'demo_form' => $form->createView(),
+            'rector_run' => $rectorRun ?? null,
```
JanMikes: Why do we here the `??`? I do not think we need it because the variable is always set.

src/Controller/DemoController.php
```diff
@@ -36,55 +36,68 @@ final class DemoController extends AbstractController
...
         ]);
     }
 
+    private function createDemoFormData(?RectorRun $rectorRun): DemoFormData
+    {
+        $formData = new DemoFormData();
+
+        if ($rectorRun) {
+            $formData->setContent($rectorRun->getContent());
```
JanMikes: I would love if it would be possible to have class `DemoFormData` without setters and only pass parameters via constructor. Can we do it?

src/Controller/DemoController.php
JanMikes: Oh my,... you just read my mind 😄 (1 comment above)

src/HttpKernel/Controller/RectorRunValueResolver.php
```diff
@@ -0,0 +1,49 @@
...
+
+    public function supports(Request $request, ArgumentMetadata $argumentMetadata): bool
+    {
+        if ($argumentMetadata->getType() !== RectorRun::class) {
+            return false;
+        }
+
+        $argumentValue = $request->get($argumentMetadata->getName());
+
+        return $argumentValue !== null;
```
JanMikes: Can we add here that `Uuid::isValid($argumentValue)` ? It would prevent InvalidUuidException

src/Utils/FileDiffCleaner.php
```diff
@@ -0,0 +1,23 @@
...
+final class FileDiffCleaner
+{
+    public function clean(string $fileDiff): string
+    {
+        // https://regex101.com/r/sI6GVY/1/
+        $match = Strings::match($fileDiff, '#^.*?@@\n(?<content>.*?)$#Us');
+        $fileDiff = $match['content'] ?? '';
+
+        if (Strings::contains($fileDiff, 'No newline at end of file')) {
+            $fileDiff = Strings::substring($fileDiff, 0, -strlen('\ No newline at end of file') - 1);
```
JanMikes: What about `Strings::before()` instead? Could be more readable than this construction

templates/homepage/demo.twig
```diff
@@ -1,32 +1,51 @@
...
-                        <a href="{{ path('demo_detail', {'id': rector_run.id }) }}">{{ rector_run.id }}</a>
-                    {% else %}
-                        ...
-                    {% endif %}
-            </div>
+            {{ form_end(demo_form) }}
         </div>
+
+        {% if rector_run is not null %}
+            <h2 class="mt-0 mb-0">Result</h2>
```
JanMikes: This is great! I hated when there was title `Result` when i did not even submitted the form 😄

config/services.yaml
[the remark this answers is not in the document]
TomasVotruba: That's what packages are for, see Blog. There is domain coupling + it works well with autodiscovery.

config/services.yaml
```diff
@@ -13,10 +13,16 @@ services:
...
 
     Symplify\PackageBuilder\Console\Style\SymfonyStyleFactory: ~
     Symfony\Component\Console\Style\SymfonyStyle:
         factory: ['@Symplify\PackageBuilder\Console\Style\SymfonyStyleFactory', 'create']
+
+    # needs to be registered manually, because it doesn't work without priority :/
+    # see https://symfony.com/doc/current/controller/argument_value_resolver.html
+    Rector\Website\HttpKernel\Controller\RectorRunValueResolver:
+        tags:
+            - { name: controller.argument_value_resolver, priority: 550 }
```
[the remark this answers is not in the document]
TomasVotruba: Everything is magic untill you read the docs :D it actually declutters the contorller code from all the nitpicking string/int id, repository, find, get, return, object work, that can also lead to bugs. 

Yes, it's for action methods only.

.gitignore
```diff
@@ -9,6 +9,7 @@ vendor/
 var
 public/bundles/
 public/build/
+public/adminer/
```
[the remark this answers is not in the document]
TomasVotruba: For local directory without Docker

src/Controller/DemoController.php
TomasVotruba: Removed in next PR

src/HttpKernel/Controller/RectorRunValueResolver.php
TomasVotruba: Not sure what exactly you mean, but sure. Send PR please.

In the end all those cases should end in 404 I guess

src/Utils/FileDiffCleaner.php
TomasVotruba: `Strings::replace()` would be even better.
Could you handle it + add test case for this class?

PR #93, opened 2020-01-19
Merge pull request #93 from rectorphp/validation-2

TomasVotruba:


src/DemoRunner.php
```diff
@@ -0,0 +1,63 @@
...
+        $stopwatch = new Stopwatch();
+        $rectorProcessStopwatchEvent = $stopwatch->start('rector-process');
+
+        try {
+            $runResult = $this->rectorProcessRunner->run($rectorRun);
+            $fileDiff = $this->createFileDiff($runResult, $rectorRun);
+
+            $rectorRun->success($fileDiff, Json::encode($runResult), $rectorProcessStopwatchEvent);
+        } catch (Throwable $throwable) {
+            $rectorRun->fail($throwable->getMessage(), $rectorProcessStopwatchEvent);
```
TomasVotruba: @JanMikes  Btw, this looks like `$rectorRun` is a service, not an object.
It would be better to use clear method naming e.g. `markAsFailed()` or create this object with a factory.

PR #94, opened 2020-01-19
Merge pull request #94 from rectorphp/bootstrap-style

TomasVotruba:
The `rector.yaml` has to be long, so it can include services names, e.g.

image

# Before

image

# After

image

### DIFF
--- a/src/ValueObject/TwigAutocomplete/DemoLink.php
+++ b/src/ValueObject/TwigAutocomplete/DemoLink.php
@@ -1,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\Website\ValueObject\TwigAutocomplete;
+
+/**
+ * Only helper Twig autocomplete class
+ */
+final class DemoLink
+{
+    /**
+     * @var string
+     */
+    public $label;
+
+    /**
+     * @var string
+     */
+    public $uuid;
+}
--- a/templates/_snippets/javascripts.twig
+++ b/templates/_snippets/javascripts.twig
@@ -59,3 +59,11 @@
         });
     });
 </script>
+
+
+<script type="text/javascript">
+    $('.btn-demo-submit').on('click', function (event) {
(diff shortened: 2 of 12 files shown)


src/ValueObject/TwigAutocomplete/DemoLink.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rector\Website\ValueObject\TwigAutocomplete;
+
+/**
+ * Only helper Twig autocomplete class
+ */
+final class DemoLink
```
JanMikes: ValueObjects should be always immutable, is it possible to make this object without public properties?

templates/_snippets/javascripts.twig
```diff
@@ -59,3 +59,11 @@
         });
     });
 </script>
+
+
+<script type="text/javascript">
+    $('.btn-demo-submit').on('click', function (event) {
+        $(this).prepend('<em class="fas fa-sync-alt fa-spin"></em>&nbsp;&nbsp;&nbsp;');
```
JanMikes: Nice one. We could as well disable the input to prevent multiple clicking.

src/ValueObject/TwigAutocomplete/DemoLink.php
TomasVotruba: Agreed. It's only for twig autocomplete of array, not real value object.

TomasVotruba: Agreed. It's only for twig autocomplete of array, not real value object.

Here: https://github.com/rectorphp/getrector.org/blob/master/templates/homepage/demo.twig#L20

Any idea how to make it work and keep code clean?

templates/_snippets/javascripts.twig
TomasVotruba: I tried many time but failed :cry: 

Feel free to add it :+1:

PR #96, opened 2020-01-19
Merge pull request #96 from rectorphp/controller-test

TomasVotruba:


src/Controller/DemoController.php
```diff
@@ -91,8 +91,6 @@ private function processFormAndReturnRoute(FormInterface $form): RedirectRespons
         $config = $demoFormData->getConfig();
 
         $rectorRun = $this->createRectorRun($config, $demoFormData);
-        $this->rectorRunRepository->save($rectorRun);
-
```
TomasVotruba: This one save() is not needed, right? It's duplicate of the line bellow

JanMikes: Yeah it is no longer required after the refactorings!

JanMikes: Yeah it is no longer required after the refactorings!

TomasVotruba: Thanks for confirmation

PR #101, opened 2020-01-20
Run demo with sudo

JanMikes:


JanMikes:
@TomasVotruba this will as well optimize and reduce docker build time by approx 70%

PR #102, opened 2020-01-20
Fix Docker security issue with sudo rm -rf

JanMikes:
As well i improved docker build speed again.

### DIFF
--- a/bin/run-demo.sh
+++ b/bin/run-demo.sh
@@ -1,14 +1,13 @@
 #!/usr/bin/env bash
 
-while getopts ":v:r:i:d:" opt
+while getopts ":v:n:i:" opt
 do
   case ${opt} in
     v) VOLUME_SOURCE=${OPTARG};;
-    r) RUN_ID=${OPTARG};;
+    n) NAME=${OPTARG};;
     i) DOCKER_IMAGE=${OPTARG};;
-    d) RUN_DIRECTORY=${OPTARG};;
     h)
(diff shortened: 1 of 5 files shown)


bin/run-demo.sh
```diff
@@ -31,24 +30,15 @@ fi
...
 
-if [[ ${RECTOR_RUN_EXIT_CODE} -eq 0 ]]
-then
-    cat ${RUN_DIRECTORY}/output.json
-else
-    docker logs ${RUN_ID}
-fi
-
-docker rm ${RUN_ID} > /dev/null
-rm -rf ${RUN_DIRECTORY}
```
JanMikes: This was the security issue. It is gone now :-) no1 wants to run `rm -rf` as sudo

src/DemoRunner.php
```diff
@@ -36,7 +36,7 @@ public function runAndPopulateRunResult(RectorRun $rectorRun): void
         $rectorProcessStopwatchEvent = $stopwatch->start('rector-process');
 
         try {
-            $runResult = $this->rectorProcessRunner->run($rectorRun);
+            $runResult = $this->rectorProcessRunner->run($rectorRun->getContent(), $rectorRun->getConfig());
```
JanMikes: @TomasVotruba i like this change very much, i was thinking about how to make it not coupled with RectorRun entity, so it could be used event without it and tested easier without mocking - this is possible now.

PR #117, opened 2020-01-23
Added ForbiddenPHPFunctionsCheckerTest

JanMikes:


JanMikes:
@TomasVotruba i think this is enough for testing checker, do we need to check it through form as well?

tests/Lint/ForbiddenPHPFunctionsCheckerTest.php
```diff
@@ -0,0 +1,28 @@
...
+        $checker = new ForbiddenPHPFunctionsChecker(['shell_exec']);
+
+        $checker->checkCode('<?php shell_exec("");');
+    }
+
+    public function testCheckShouldPass(): void
+    {
+        $this->expectNotToPerformAssertions();
+
+        $checker = new ForbiddenPHPFunctionsChecker(['shell_exec']);
```
TomasVotruba: Should be in `setUp()` at least

src/Lint/ForbiddenPHPFunctionsChecker.php
```diff
@@ -24,8 +24,8 @@ public function __construct(array $forbiddenFunctions)
 
     public function checkCode(string $code): void
     {
-        // https://regex101.com/r/4in3xJ/2
-        $pattern = sprintf('#^(?<function>%s)\s*\(#mi', implode('|', $this->forbiddenFunctions));
+        // https://regex101.com/r/4in3xJ/3
```
TomasVotruba: :cupid:

tests/Lint/ForbiddenPHPFunctionsCheckerTest.php
TomasVotruba: Better use service, so we use and test current setup.
`shell_exec` might be here, but not in the params, which would fuckup

JanMikes: I disagree.

Definitely (unit) tested class should be created in the test method and not in setup. It is possible to decouple it to private method `createChecker()` though.

I am big fan of unit tests as they can be used in TDD very easily. 

I believe it is ok to have class covered by tests twice, once unit and once in integration, what do you think?

TomasVotruba: This is not about theory, it's about added value. Test should help. 
This test only tests passing string, which is useless.
The app can be actually broken and form deleted.

JanMikes: > This test only tests passing string, which is useless.

Nope, it actually tests 2 scenarios :-). Just imagine the situation, you will want to improve/change how detection of forbidden function works and you will extend DemoControllerTest for that? It seems as very bad practice for me.

> That's pointless. Use just one that does the coverage for app.

I see difference in that. Unit test can help me during feature development, i test that the class work as expected. In integration tests i test that the class is integrated correctly and works together with the system. 

I still believe keeping both tests is the best way as both of them will test completely different things, yet they will cover same class twice.

TomasVotruba: In theory yes, in practise it only adds complexity and maintenance overhead. See Ocamm's Razor.

Same goes for DemoControllerTest. It doesn't make sense to duplicate it's testing features.

This test is not practical, as it exposes possible bug.

TomasVotruba:
Conflicting and needs rebase. I'll resolve this locally

PR #128, opened 2020-02-01
Remove dead line

### LINKED ISSUE #124: Demo removes <?php line
Live demo: https://getrector.org/demo/97209fdb-db60-4425-9278-ca2c00e7836c#result

PHP snippet:
```php
<?php declare(strict_types=1);

class CommonDemoFile
{
	public function run(array $params)
    {

    }
}

class DemoFile extends CommonDemoFile
{
    public function run(array $params)
    {
        return $params;
    }
}
```

```yml
# rector.yaml
parameters:
    sets:
        - 'dead-code'
```

Unexpected: Rector removes `<?php declare(strict_types=1);`

TomasVotruba:
Closes #124


src/FormDataFactory/DemoFormDataFactory.php
```diff
@@ -0,0 +1,25 @@
...
+final class DemoFormDataFactory
+{
+    public function createFromRectorRun(?RectorRun $rectorRun): DemoFormData
+    {
+        if ($rectorRun) {
+            return new DemoFormData($rectorRun->getContent(), $rectorRun->getConfig());
+        }
+
+        // default values
+        $demoContent = FileSystem::read(__DIR__ . '/../../data/DemoFile.php');
```
JanMikes: Ultimately these paths could be passed as parameters via constructor, so it is easier to test this class.

tests/FormDataFactory/DemoFormDataFactoryTest.php
```diff
@@ -0,0 +1,46 @@
...
+final class DemoFormDataFactoryTest extends AbstractKernelTestCase
+{
+    /**
+     * @var DemoFormDataFactory
+     */
+    private $demoFormDataFactory;
+
+    protected function setUp(): void
+    {
+        self::bootKernel(GetRectorKernel::class);
```
JanMikes: `$this->bootKernel()` should be used instead, `bootKernel()` method is not static. PHPStorm is reporting this to me as bug.

JanMikes: I see that symfony's `KernelTestCase` is static but in symplify `AbstractKernelTestCase` it is not static

src/FormDataFactory/DemoFormDataFactory.php
TomasVotruba: Parameters suggest it's configurable, which is not.
Also `FileSystem` is 3rd party code, we don't need to test it.

Might be constants though to make it more clear :+1: Thanks for the hint

tests/FormDataFactory/DemoFormDataFactoryTest.php
TomasVotruba: Yes indeed :+1: Thanks

src/FormDataFactory/DemoFormDataFactory.php
TomasVotruba: Also, parameters or compositive paths (`$someDir . '/path.php') suck at file jumping:

jump

public/assets/prism/prism.js
```diff
@@ -5,7 +5,7 @@ Prism.languages.markup={comment:/<!--[\s\S]*?-->/,prolog:/<\?[\s\S]+?\?>/,doctyp
...
-Prism.languages.diff={coord:[/^(?:\*{3}|-{3}|\+{3}).*$/m,/^@@.*@@$/m,/^\d+.*$/m],deleted:/^[-<].*$/m,inserted:/^[+>].*$/m,diff:{pattern:/^!(?!!).+$/m,alias:"important"}};
+Prism.languages.diff={coord:[/^(?:\*{3}|-{3}|\+{3}).*$/m,/^@@.*@@$/m,/^\d+.*$/m],deleted:/^[-].*$/m,inserted:/^[+>].*$/m,diff:{pattern:/^!(?!!).+$/m,alias:"important"}};
```
TomasVotruba: This took me 30 mins to discover, terrible :D

```diff
- remove
-?php remove

< remove
<?php remove
```

PR #173, opened 2020-09-03
Add logo variant featuring both text and the house

honzajavorek:
I downloaded the Montserrat font and tried to reproduce the logo as it is on the website, but in Inkscape so it can be featured as a standalone SVG logo.


honzajavorek:
Hmm, in the preview here I noticed the SVG isn't cropped for some reason. I'll try to fix that.

TomasVotruba:
Wohoo, thank you for this work!

Not sure if I'm seeing it right. Maybe I miss some local fonts...

Here:

image


On website:

image

honzajavorek:
Huh, I see it right, probably because I have the font installed now :) I’ll
see if I can force Inkscape to include the letters as curves instead, so
it's independent on system fonts.

TomasVotruba:
If it's a problem to include the font and it helps you solve the missing logo, I can merge it.

honzajavorek:
Dont merge yet, I’ll update the PR with the letters as curves. It must have
solution, its a common practice. I just thought Inkscape did it
automatically as I couldn’t notice it did not :))

TomasVotruba:
All right, just ping me when it's ready :)

honzajavorek:
Does it display better now?

TomasVotruba:
Better :+1: 

https://github.com/rectorphp/getrector.org/pull/173/files?short_path=8b10bce#diff-8b10bcefa798e66cf68ef789312f007a

image


:clap: 

How difficult is to make also white variant? So we can use it instead of current text-picture mixed logo in the future

honzajavorek:
I think it should be easy-ish, another 15mins of work at maximum.

honzajavorek:
Could you verify if works correctly? (As you don't have Montserrat font installed, I'd have to uninstall it first to verify.)

TomasVotruba:
Pufrect! Now we're a serious company :), thanks :+1: 

image

honzajavorek:
🚀

PR #177, opened 2020-09-10
Hotfix missing version in rector run

JanMikes:
Fixes GETRECTOR-ORG-R


TomasVotruba (commented):
This seems like bug in Rector.

Maybe caused by this?
https://github.com/rectorphp/rector/pull/4141

TomasVotruba:
Thanks for the fix

JanMikes:
@TomasVotruba i can confirm - caused by https://github.com/rectorphp/rector/commit/a32b4fed64af7ac27b2b5bed92b4bd29b33dd79b

PR #187, opened 2020-11-22
Fix the count

orklah:
This should fix https://github.com/rectorphp/getrector.org/issues/186

However, I didn't test it at all because I was not sure how

I also changed the fallback count to 600? The real count is 625 right now if my regex is correct


TomasVotruba:
Looks good, thanks!

TomasVotruba:
Thanks to this issues I've finally added category support to @symplify - https://github.com/symplify/symplify/pull/2547

Now should be restored to almost original format:
https://github.com/rectorphp/rector/pull/4688

PR #190, opened 2020-12-02
Upgrade to PHP 8

TomasVotruba:


Dockerfile
```diff
@@ -83,7 +83,7 @@ FROM base as production
 
 COPY composer.json composer.lock phpunit.xml ./
 
-RUN composer install --prefer-dist --no-dev --no-autoloader --no-scripts --no-progress \
+RUN composer install --prefer-dist --no-dev --no-autoloader --no-scripts --no-progress --ignore-platform-reqs \
     && composer clear-cache
 
 COPY --from=node-build /build/public/build ./public/build
```
TomasVotruba, line 89: Hi @JanMikes ,

I'm finally doing upgrade to PHP 8 :). 
I've have only changed the base image from 7.4 to 8.

Just in case it needs some upgrade on the server, could you handle it?

JanMikes: Hi, im available tonight. If anything goes wrong, pm me ;).

TomasVotruba: Cool :) I'm gonna merge it in 2 minutes, so the website should be down at 5 mins :D

PR #213, opened 2020-12-13
improve Docker for local dev

TomasVotruba:
Mirror to https://github.com/deprecated-packages/cleaningladylist.com/pull/24


.github/workflows/deploy.yaml
```diff
@@ -25,27 +25,8 @@ jobs:
...
-
-                docker build . --tag rector/getrector.org:latest --target production \
-                  --build-arg commitHash=${{ github.sha }} \
-                  --cache-from rector/getrector.org:base \
-                  --cache-from rector/getrector.org:node-build \
-                  --cache-from rector/getrector.org:latest
-
-                docker push rector/getrector.org:base
-                docker push rector/getrector.org:node-build
+                docker build . --tag image --cache-from rector/getrector.com:latest
```
TomasVotruba, line 29: @JanMikes I tried to mimic https://github.com/deprecated-packages/cleaningladylist.com/pull/24/files#diff-4f9e38227ed64fefb17f4668a7ac4ab55b6149994d5ac2fd96182d5958479b54
here. Making it simple :) 

What alse should be fixed here? :)

.github/workflows/deploy.yaml
```diff
@@ -25,27 +25,8 @@ jobs:
...
-
-                docker build . --tag rector/getrector.org:latest --target production \
-                  --build-arg commitHash=${{ github.sha }} \
-                  --cache-from rector/getrector.org:base \
-                  --cache-from rector/getrector.org:node-build \
-                  --cache-from rector/getrector.org:latest
-
-                docker push rector/getrector.org:base
-                docker push rector/getrector.org:node-build
+                docker build . --tag rector/getrector.com:latest --cache-from rector/getrector.com:latest
```
TomasVotruba, line 29: @JanMikes I tried to mimic https://github.com/deprecated-packages/cleaningladylist.com/pull/24/files#diff-4f9e38227ed64fefb17f4668a7ac4ab55b6149994d5ac2fd96182d5958479b54
here. Making it simple :) 

What alse should be fixed here? :)

TomasVotruba:
The CI is green, I'll try to ship it :)

.github/workflows/deploy.yaml
JanMikes: well, it was performance optimization here, it will work anyways even without it.

TomasVotruba: I see. The build job is done in 3,5 mins, which is great. Even 5 would be ok.



Simpler scripts makes it easier to maintain and upgrade (e.g. now to PHP 8), so it's positive benefits to have fewer lines.

PR #246, opened 2021-02-01
Show fatal error message with newlines

ruudk:
## Old
Screenshot 2021-02-06 at 10 04 53@2x

## New
Screenshot 2021-02-06 at 10 05 03@2x

### DIFF
--- a/templates/demo/demo.twig
+++ b/templates/demo/demo.twig
@@ -81,7 +81,7 @@
         {% if rector_run.hasRun and rector_run.successful != true %}
             <div class="alert alert-danger">
                 <strong>Fatal error:</strong>
-                {{ rector_run.fatalErrorMessage }}
+                {{ rector_run.fatalErrorMessage | nl2br }}
             </div>
         {% endif %}
 


TomasVotruba:
Is it before or after?
How does the other version look?

ruudk:
This is before. I don't know how after looks like 😅

TomasVotruba:
It must be tested before random pushing to production :)
You should be able to see it on Docker run

ruudk:
I'll do that later this week. Will also investigate the issue with the capped diff output (other issue).

ruudk:
I'm having difficulties running it locally.

Why do I need to both start `docker-compose` (that seems to run web, js, mysql) + run `npm watch`, `php server` etc... ? Can't I run it with just a single `docker-compose up`?

ruudk:
Are you sure the Readme steps are still up to date?

ruudk:
When running  `php -S localhost:8000 -t public` I get the following error when I try to Process some code on the Demo:
```
An exception occurred in driver: SQLSTATE[HY000] [2002] No such file or directory
```

JanMikes:
Hi, you are right, readme is not up to date. Just run `bin/run-from-docker.sh` and you should be okay, as you noticed it runs everything, including npm watcher. Site is then accessible on `localhost:8080`

Thank you for letting us now, i will take care of updating readme. In case you struggle with running project locally, please open new issue and i will be happy to help you as i am author of the docker setup.

JanMikes:
> An exception occurred in driver: SQLSTATE[HY000] [2002] No such file or directory

If you run in docker, it might be related to this commit https://github.com/rectorphp/getrector.org/commit/ba72e9132f76a35afa423ca75c41c297cad3d1c6#diff-57248549079ffc1cf256bcae26783b17ebc4a970b390528f781bcbe20a55531eR36

For some reason tomas changed mysql port to 3336 (i think he had a collision with other locally running projects, but i believe it should be done in other way, does not matter anyways).

ruudk:
@TomasVotruba Updated the PR description with screenshots. Much better, isn't it?

TomasVotruba:
That's great, thank you 🤗

Issue #245, opened 2021-01-30
[demo] Output is capped when multiple classes/interfaces


ruudk:
See https://getrector.org/demo/3988c72f-7a5b-4f9d-ba87-1eb209e6d1d5

```php
<?php
interface FormTypeInterface
{
    public function getTag() : string;
}

final class DemoFile
{
    /**
     * @var iterable|FormTypeInterface[]
     */
    private $formTypes;

    /**
     * @param FormTypeInterface[] $formTypes
     */
    public function __construct(iterable $formTypes)
    {
        $this->formTypes = $formTypes;
    }
}
```

shows like this:
<img width="1001" alt="Screenshot 2021-01-30 at 13 47 10@2x" src="https://user-images.githubusercontent.com/104180/106356662-ab00b700-6301-11eb-9760-83dba79b3ed2.png">

TomasVotruba:
Test would be great to cover this :+1:

ruudk:
See https://github.com/rectorphp/getrector.org/pull/253

TomasVotruba:
Closed by https://github.com/rectorphp/getrector.org/pull/253

Tip for next time automation:
https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

PR #253, opened 2021-02-06
Multiple classes diff is wrong

ruudk:
See #245

I tried to fix this but it's not as easy. The regex always looks for the last part of the diff.

Wouldn't it be better to just show the full diff of the file? Or just show the full content of the changed file? Maybe side by side?

### DIFF
--- a/packages/demo/src/Utils/FileDiffCleaner.php
+++ b/packages/demo/src/Utils/FileDiffCleaner.php
@@ -13,10 +13,16 @@ use Rector\Website\Demo\Entity\RectorRun;
 final class FileDiffCleaner
 {
     /**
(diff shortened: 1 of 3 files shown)


TomasVotruba:
We can change to one of your suggestions. Maybe full diff would work. Screenshot would help to decide

ruudk:
This is how it looks when there is only 1 diff part:
Screenshot 2021-02-15 at 19 18 24@2x


This is how it looks for when there are multiple diff parts:
Screenshot 2021-02-15 at 19 17 51@2x


@TomasVotruba WDYT?

TomasVotruba:
Looks good :+1: Let's ship it

PR #254, opened 2021-02-06
Resolve directory for demo temp files on host dynamically

### LINKED ISSUE #249: [demo] Running locally gives Fatal error: [ERROR] File "/project/rector.php" was not found
I'm using `bin/run-from-docker.sh` to start the project, then I go to http://localhost:8080/ and visit the website. I click demo and click `Process`. then I get:

Screenshot 2021-02-03 at 18 55 41@2x

Would be really great if the Readme can be fixed + tested on a new checkout so that everything works.

JanMikes:
Fixes #249

### DIFF
--- a/.env
+++ b/.env
@@ -24,7 +24,7 @@ DATABASE_CHARSET=utf8
 
 DEBUG_COOKIE=
 
-HOST_DEMO_DIR=/var/www/getrector.org/var/demo
+HOST_DEMO_DIR=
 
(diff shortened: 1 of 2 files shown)


TomasVotruba:
I guess this causes the same bug on Linux :D

JanMikes:
@TomasVotruba pls share more info. It works on MacOS, Ubuntu and either on production, this must be something with your machine.

PR #251, opened 2021-02-05
Update docker setup

JanMikes:

PR #262, opened 2021-02-10
Revert "Update docker setup (#251)"

TomasVotruba:
This reverts commit 4a03bd7a249eae6b79b7a4f3716cb97f4fabac2e.

Hotfix for 2 Docker bugs on Linux:

- https://github.com/rectorphp/getrector.org/pull/251#issuecomment-777072338
- https://github.com/rectorphp/getrector.org/pull/254#issuecomment-777079386

Feel free to revert when a propper fix is ready. I need to make some fixes on demo form now, so I need working Docker form

### DIFF
--- a/.docker/docker-entrypoint
+++ b/.docker/docker-entrypoint
@@ -1,0 +1,33 @@
+#!/bin/bash
+set -e
+
+# first arg is `-f` or `--some-option`
+if [ "${1#-}" != "$1" ]; then
+	set -- apache2-foreground "$@"
+fi
+
+if [ "$1" = 'apache2-foreground' ] || [ "$1" = 'bin/console' ] || [ "$1" = 'php' ] || [ "$1" = 'composer' ]; then
+	if [ "$APP_ENV" != 'prod' ]; then
+		composer install --prefer-dist --no-progress --no-interaction -o --ignore-platform-req php
+	fi
+
+    php bin/console assets:install
+    php bin/console cache:clear
+
+    ## Check if variable DATABASE_HOST is set, if yes, we have database
+    if [[ -v DATABASE_HOST ]]; then
+        ## Wait until database connection is ready
+        until mysql -u $DATABASE_USER -h $DATABASE_HOST --password="$DATABASE_PASSWORD" -e "" ; do
+            >&2 echo "Waiting for database service to start."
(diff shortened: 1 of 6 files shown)


JanMikes:
Pls no panic revert, im here to help you (can you give it few mins?). As it works on production and tested on 2 different OS (macos, ubuntu) i suspect there is something with your machine.

Please: rm -rf repository, git clone and `docker-compose up` -> do fresh startup, tested 5 times now on ubuntu.

JanMikes:
FIrst guess, think it could be cased by `.env.local` existing in your repository, check if it exists and if yes, delete it as well in case you dont want to do fresh startup.

TomasVotruba:
I'll give it try after the fix

TomasVotruba:
I'll give it try after the fix

TomasVotruba:
I'll give it try after the fix

JanMikes:
Nice! 👌 im here to help you in realtime, on any channel, whatsapp, telegram, messenger, sms, choose your favorite 😸

ruudk:
@TomasVotruba Are you running Docker (compose) with sudo by any chance? 

See https://stackoverflow.com/a/61905266
https://docs.docker.com/engine/install/linux-postinstall/

That could explain why `$PWD` is empty. 

I think we should try to fix it like it was, because then it would just work by default, without having to specify any paths.

TomasVotruba:
@ruudk Yes, always. Otherwise I get these errors:

https://github.com/symplify/symplify/blob/1353bd6ae212d520f7b7a43902e83edae8c91bcb/tests/bootstrap.php#L21

ruudk:
@TomasVotruba wrong link? It doesn't show the errors

TomasVotruba:
@ruudk I'm sorry, I got confused with another issue. I've updated the image

ruudk:
https://stackoverflow.com/a/52798075  Seems like your `/var/run/docker.sock` needs a chown.

ruudk:
Or follow these steps, https://docs.docker.com/engine/install/linux-postinstall/ 

You should not have to run it with sudo.

TomasVotruba:
I would need exact steps, as I don't have the skill to work with bash and user access rights.

PR #263, opened 2021-02-10
Fix falsy report on demo diff

TomasVotruba:
Fixes https://github.com/rectorphp/getrector.org/issues/220


docker-compose.yml
```diff
@@ -34,7 +34,7 @@ services:
         volumes:
             - ./.docker/mysql-data:/var/lib/mysql
         ports:
-            - "${GETRECTOR_ORG_MYSQL_PORT:-3306}:3306"
+            - "${GETRECTOR_ORG_MYSQL_PORT:-3307}:3306"
```
TomasVotruba, line 37: This extra step caused me annoying delay that might have lead to revert.
Better keep it simple as before to avoid troubles :)

JanMikes: But why? If you need sone dofferent port than the default, use env var for it. It is documented and it is convention. This change might be confusing and might cause annonying delay to anyone else. It was feedback based change. I know you were used to it but isnt better to teach you the right way instead of forcing others to use something else?

JanMikes: Overall it is not a big deal, im just curious as it changes sone common defaults :-)

TomasVotruba: It's extra energy for no gain. I consider Docker setup to ease the work, so anything extra is contra the Docker idea :)

PR #268, opened 2021-02-12
[Demo] Process directly during HTTP request

JanMikes:
Goal is to create server that will over TCP accept content of file + rector config, process it via rector and return results as json. 

This solution is supposed to replace current demo architecture.


JanMikes:
@TomasVotruba previously code like this crashed rector demo on timelimit and meanwhile running (60s before killed) it took all CPU resources on the server:
Screenshot 2021-03-03 at 0 24 48
Screenshot 2021-03-03 at 0 07 21

Now:  
Screenshot 2021-03-03 at 0 07 41

With under 100ms with xdebug enabled and debug mode 😄   
Screenshot 2021-03-03 at 0 08 10

bin/run-demo.sh
```diff
@@ -1,47 +0,0 @@
...
-shift $((OPTIND - 1))
-
-docker run \
-    --name ${NAME} \
-    --volume ${VOLUME_SOURCE}:/project \
-    --env COLUMNS=110 \
-    --env LINES=100 \
-    ${DOCKER_IMAGE} \
-    process /project/rector_analyzed_file.php \
-    --autoload-file /project/rector_analyzed_file.php \
```
TomasVotruba, line 39: This can now removed, right?

JanMikes: Yep, deleting this was not an accident :-)

TomasVotruba: That would be nice accident :D I meant the `--autoload-file` option :)

With static reflection, the autoload file does not seems useful anymore

JanMikes: I see, anyways it is no longer used anywhere so all good 👌

Issue #270, opened 2021-02-14
[demo] broken


ruudk:
See https://getrector.org/demo/792bd08a-4650-4f13-a5ac-b3135a4837ef bottom 

```
Fatal error: Fatal error: Uncaught Error: Call to undefined function Symfony\Component\Console\proc_open() in /rector/vendor/symfony/console/Terminal.php:162
Stack trace:
#0 /rector/vendor/symfony/console/Terminal.php(148): Symfony\Component\Console\Terminal::readFromProcess('stty -a | grep ...')
#1 /rector/vendor/symfony/console/Terminal.php(114): Symfony\Component\Console\Terminal::getSttyColumns()
#2 /rector/vendor/symfony/console/Terminal.php(97): Symfony\Component\Console\Terminal::initDimensionsUsingStty()
#3 /rector/vendor/symfony/console/Terminal.php(33): Symfony\Component\Console\Terminal::initDimensions()
#4 /rector/vendor/symfony/console/Style/SymfonyStyle.php(51): Symfony\Component\Console\Terminal->getWidth()
#5 /rector/src/Console/Style/SymfonyStyleFactory.php(40): Symfony\Component\Console\Style\SymfonyStyle->__construct(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#6 /rector/bin/rector.php(39): Rector\Core\Console\Style\SymfonyStyleFactory->create()
#7 /rector/bin/rector(4): require_once('/rector/bin/rec...')
#8 {main}
thrown in /rector/vendor/symfony/console/Terminal.php on line 162

```

@JanMikes

JanMikes:
Hi, im on it, might be related to yesterday update of demo to run with php 8

JanMikes:
@ruudk it works now 😉 

https://getrector.org/demo/265c6e32-c321-4fe3-8dd1-859eb1c4f8f2

ruudk:
Thanks for fixing!

PR #273, opened 2021-02-15
Demo > Show failure always on top

ruudk:
When working on small screens, it's very easy to mis the error messages as they are located under the Process button.

This is also why it took me some time to know that the Demo was broken (#270).

This will improve it.

Screenshot 2021-02-15 at 11 56 17@2x

### DIFF
--- a/templates/demo/demo.twig
+++ b/templates/demo/demo.twig
@@ -6,6 +6,15 @@
     <div id="rector_run_form" class="mt-4 mb-5">
         {{ form_start(demo_form) }}
 
+        {# @var rector_run \Rector\Website\Demo\Entity\RectorRun #}
+
+        {% if rector_run.hasRun and rector_run.successful != true %}
+            <div class="alert alert-danger">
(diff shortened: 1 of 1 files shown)


TomasVotruba:
Looks good!


Could you remove the extra space under the text? 
Demo-Show-failure-always-on-top-by-ruudk-·-Pull-Request-273-·-rectorphp-getrector-org

So it looks esteticaly

ruudk:
I did not change the styling of the error message. For the screenshot I removed the original error and replaced it with this text because I did not wanted to share the real error message I was getting (pssssst: #249)

ruudk:
Ok, now that I got the demo working again, this is how it look:
Screenshot 2021-02-15 at 16 11 12@2x

Added a trim, so it looks like:
Screenshot 2021-02-15 at 16 11 49@2x


Better :) ?

TomasVotruba:
Purfect, thank you :hugs:

Issue #247, opened 2021-02-03
[blog] add socials and rss icons


TomasVotruba:
A tip from @phpfour 

![socials](https://user-images.githubusercontent.com/924196/106755337-4f4e6a80-662e-11eb-9c86-9eb3ce2bd3ce.png)

![unnamed](https://user-images.githubusercontent.com/924196/106755513-81f86300-662e-11eb-852f-e9f27c931daa.png)

PR #276, opened 2021-02-17
[blog] add rss icon on header

phpfour:
Adds the RSS icon right next to the header for a better discovery (#247).

image


TomasVotruba:
Hi, thanks for the PR!

The typical color for RSS is orange, so people might find it faster. Could you make it orange here?
Thanks

PR #286, opened 2021-03-06
[Rector Config] Add Option::PHPSTAN_FOR_RECTOR_PATH to rector config

samsonasik:


TomasVotruba:
What is the motivation for this?

samsonasik:
I checked the patch of https://github.com/rectorphp/rector/pull/5781 seems not applied in getrector.org/demo , is it maybe probably by this or because of pending update in the getrector.org site ?

TomasVotruba:
Website demo is not deployed as result of merge here.

Last was 3 days ago: https://github.com/rectorphp/getrector.org

samsonasik:
Ok :+1:

PR #288, opened 2021-03-14
Add RSS link on blog page

TomasVotruba:
Replaces https://github.com/rectorphp/getrector.org/pull/276

Ref https://github.com/rectorphp/getrector.org/issues/247


TomasVotruba:
image

PR #295, opened 2021-03-20
Fix missing version

TomasVotruba:
Fixes https://github.com/rectorphp/getrector.org/issues/293


TomasVotruba:
I tried to fix some other bug in `rector/rector` 2 days ago, but it broke the versoin plugin.
This fixes it https://github.com/rectorphp/rector/commit/49c81bec47c7781e934eadbed85569047d156847

PR #305, opened 2021-04-22
Create automated_updates.yaml

samsonasik:
Based on https://github.com/rectorphp/rector/pull/6197#issuecomment-824789639


TomasVotruba:
Could we make it part of existing build to avoid duplication https://github.com/rectorphp/getrector.org/blob/main/.github/workflows/production_build.yaml ?

samsonasik:
@TomasVotruba implemented https://github.com/rectorphp/getrector.org/pull/305/commits/91315ec53bd9343d0317c6a66bde7f7a376dfbdd

.github/workflows/production_build.yaml
```diff
@@ -1,6 +1,10 @@
 name: Production build
 
-on: push
+on:
+  push: null
+  schedule:
+        # https://crontab.guru/every-7-days
+        -   cron: "0 0 * * 0"
```
TomasVotruba, line 7:
```suggestion
        # https://crontab.guru/daily
        -   cron: "0 0 * * *"
```

.github/workflows/production_build.yaml
```diff
@@ -1,6 +1,10 @@
 name: Production build
 
-on: push
+on:
+  push: null
+  schedule:
+        # https://crontab.guru/daily
+        -   cron: "0 0 * * *"
```
TomasVotruba, line 7:
```suggestion
on:
    push: null
    schedule:
        # https://crontab.guru/daily
        -   cron: "0 0 * * *"
```

TomasVotruba: space indent

TomasVotruba (approved):

TomasVotruba:
Thanks :+1:

JanMikes:
Wow, this is easy and clever hotfix, did not realise it can be done in such easy way, i was thinking about triggers from rector repo to this repo to trigger build 😄 good job

JanMikes:
But w8, this pull request does not solve updating, this will just re-deploy the same version of website that is already running, WITHOUT updated rector dependency. 

You need to run `composer update` before and changes in composer.lock should be commited

TomasVotruba:
cc @samsonasik

samsonasik:
For new commit of composer.lock, it seems needs git diff first to check if there is any file change?

PR #308, opened 2021-04-24
[GHA] Revert auto-commit

samsonasik:
ref https://github.com/rectorphp/getrector.org/pull/307#issuecomment-826102050


samsonasik:
@TomasVotruba reverted

TomasVotruba:
Thanks

PR #310, opened 2021-05-06
[GHA] Fix action/checkout github token

samsonasik:


samsonasik:
@TomasVotruba the current daily pull request automatic update got error:

```bash
Run actions/checkout@v2

Error: Input required and not supplied: token
```

ref https://github.com/rectorphp/getrector.org/runs/2505391808?check_suite_focus=true#step:2:1

It may be because `${{ secrets.ACCESS_TOKEN }}` not exists, could you verify if it is should be `${{ secrets.ACCESS_TOKEN }}` or `{{ $secrets.GITHUB_TOKEN }}` in this repo?

TomasVotruba:
Hi, I've checked it now and added `ACCESS_TOKEN`

TomasVotruba:
I've just retriggered the workflow

samsonasik:
@TomasVotruba thank you, it seems working now https://github.com/rectorphp/getrector.org/pull/317 , I am closing it.

PR #328, opened 2021-05-22
[GHA] Per-3 days automatic update rector/rector only dependency

samsonasik:
Fixes #327  refs https://github.com/rectorphp/getrector.org/pull/327#issuecomment-846382616


samsonasik:

PR #339, opened 2021-05-29
Apply monthly schedule from dependabot

samsonasik:
ref https://github.com/rectorphp/getrector.org/pull/338#issuecomment-850911245 Closes https://github.com/rectorphp/getrector.org/pull/338


samsonasik:

samsonasik:
@TomasVotruba let's try this, if it works, next update schedule is first day of the month.

PR #386, opened 2021-06-23
use access token for daily create pull request

samsonasik:
to make github action running ref https://github.com/rectorphp/getrector.org/pull/385#issuecomment-866421904


samsonasik:
Let's give it a try, @TomasVotruba could you merge it as I don't have acces to merge.

TomasVotruba:
I think these 2 tokens are aliases.

PR #387, opened 2021-06-23
Specify branch main on production_build.yaml and unit_tests.yaml

samsonasik:
To avoid production build and Unit test gha hanging:

Screen Shot 2021-06-23 at 12 40 12

### DIFF
--- a/.github/workflows/production_build.yaml
+++ b/.github/workflows/production_build.yaml
@@ -1,6 +1,7 @@
 name: Production build
 
 on:
+    pull_request: null
     push:
         branches:
             - main
--- a/.github/workflows/production_build.yaml
+++ b/.github/workflows/production_build.yaml
@@ -1,6 +1,9 @@
 name: Production build
 
-on: push
+on:
+    push:
+        branches:
+            - main
 
 jobs:
     warmup:
--- a/.github/workflows/unit_tests.yaml
+++ b/.github/workflows/unit_tests.yaml
@@ -1,6 +1,9 @@
 name: Unit Tests
 
-on: push
+on:
+    push:
+        branches:
+            - main
 
 jobs:
     unit_tests:
--- a/.github/workflows/unit_tests.yaml
+++ b/.github/workflows/unit_tests.yaml
@@ -1,6 +1,7 @@
 name: Unit Tests
 
 on:
+    pull_request: null
     push:
         branches:
             - main


samsonasik:
It is strange, the "Cache warmup" and "unit_tests" keep showing "Waiting for status to be reported"

TomasVotruba:
That's because those are "required" before merge.

Unit test must be invoked too:
https://github.com/rectorphp/getrector.org/blob/main/.github/workflows/unit_tests.yaml

samsonasik:
How to disable them as they only on push, not pull request ?

TomasVotruba:
> How to disable them as they only on push, not pull request ?

I would enable them on push too. Having passed unit test is crutial.

I've added this repository to your access. You should be able to merge now

samsonasik:
They were only on push, do you mean they need to be run in "pull_requests" as well? For unit tests, that's ok, for "Cache warmup", that seems change production docker?

samsonasik:
I enabled "Unit tests" for running on both "pull request" and "push", @TomasVotruba could you disable "Cache warmup" action from required in pull request, as it seems push to production?

TomasVotruba:
Cache warmup must pass too. If it fails, the production might be pushed and broken (this happened yestereday due to Symfony renamed class BC break)

samsonasik:
Ok, I enable "Cache warmup" to run on pull request too.

.github/workflows/production_build.yaml
```diff
@@ -1,6 +1,10 @@
 name: Production build
```
TomasVotruba, line 1: @samsonasik It seems I mixed the naming here. Could you update the file name + name to "cache_warmup"? It's definitely not a production build :)

.github/workflows/production_build.yaml
```diff
@@ -1,6 +1,10 @@
...
 
-on: push
+on:
+    pull_request: null
+    push:
+        branches:
+            - main
```
TomasVotruba, line 7: :+1:

.github/workflows/production_build.yaml
samsonasik: implemented https://github.com/rectorphp/getrector.org/pull/388

TomasVotruba: Thanks! :slightly_smiling_face:  Now the confusion makes sense, I've picked very bad naming :/

PR #400, opened 2021-07-06
add abdul and sebastian to team

TomasVotruba:


samsonasik:
I updated composer.lock to use`rector/rector` to 0.11.35

PR #408, opened 2021-07-24
[Dependencies] Update symplify/* and phpstan dependency

samsonasik:
- update symplify/* packages to latest `^9.4.23`
- update phpstan to use `^`:

```diff
-"phpstan/phpstan": "0.12.91",
+"phpstan/phpstan": "^0.12.91",
```

as the latest rector now using pinned phpstan version and latest `dev-main` using pinned `0.12.93` : 

https://github.com/rectorphp/rector/blob/9ffc8ec64a8bda1e078ec68f75b246795a693be8/composer.json#L10

so it is safe to use `^` and updating rector.


composer.json
```diff
@@ -43,32 +43,32 @@
...
-        "symplify/markdown-diff": "9.4.22",
-        "symplify/package-builder": "9.4.22",
-        "symplify/php-config-printer": "9.4.22",
-        "symplify/rule-doc-generator": "9.4.22",
-        "symplify/simple-php-doc-parser": "9.4.22",
-        "symplify/skipper": "9.4.22",
-        "symplify/smart-file-system": "9.4.22",
-        "symplify/symfony-php-config": "9.4.22",
-        "symplify/symplify-kernel": "9.4.22"
+        "symplify/amnesia": "^9.4.23",
```
samsonasik, line 46: @TomasVotruba do you think it should be `9.4.23` or `^9.4.23` ?

composer.json
```diff
@@ -43,32 +43,32 @@
...
+        "symplify/rule-doc-generator": "^9.4.23",
+        "symplify/simple-php-doc-parser": "^9.4.23",
+        "symplify/skipper": "^9.4.23",
+        "symplify/smart-file-system": "^9.4.23",
+        "symplify/symfony-php-config": "^9.4.23",
+        "symplify/symplify-kernel": "^9.4.23"
     },
     "require-dev": {
-        "phpstan/phpstan": "0.12.91",
+        "phpstan/phpstan": "^0.12.91",
```
samsonasik, line 61: using this to avoid overlap rector cannot update as it pinned phpstan 0.12.91 while latest rector uses phpstan 0.12.93

composer.json
TomasVotruba: Definitely `^9.4`, I was probably testing something and forgot to open the dependency up

composer.json
TomasVotruba: I think we can go to 0.12.93 here. I've upgraded the Rector to use the latest PHPStan

samsonasik: that's not work as current rector still uses phpstan 0.12.91, update to use `0.12.93` will get error:

```bash
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - rector/rector[0.11.36, ..., 0.11.38] require phpstan/phpstan 0.12.91 -> found phpstan/phpstan[0.12.91] but it conflicts with your root composer.json require (^0.12.93).
    - Root composer.json requires rector/rector ^0.11.36 -> satisfiable by rector/rector[0.11.36, 0.11.37, 0.11.38].

Use the option --with-all-dependencies (-W) to allow upgrades, downgrades and removals for packages currently locked to specific versions.
```

TomasVotruba: It was not released yet probably.
Could you tag rector if you have access? I'm not on PC now

samsonasik: Ok, I just tagged 0.11.39 https://github.com/rectorphp/rector-src/releases/tag/0.11.39

samsonasik: updated to use phpstan 0.12.93 and rector to 0.11.39

PR #441, opened 2021-11-06
[Daily] Fix daily update composer update

samsonasik:
Update `"phpstan/phpstan-doctrine` to `^1.0` and `symplify` packages to `^10.0`.

solve https://github.com/rectorphp/getrector.org/runs/4122986149?check_suite_focus=true#step:4:12


packages/Demo/Entity/RectorRun.php
```diff
@@ -18,7 +18,7 @@
 use Rector\Website\Exception\ShouldNotHappenException;
 use Symfony\Bridge\Doctrine\IdGenerator\UuidGenerator;
 use Symfony\Component\Uid\Uuid;
-use Symplify\SymplifyKernel\Strings\StringsConverter;
+use Symplify\PackageBuilder\Strings\StringFormatConverter;
```
samsonasik, line 21: @TomasVotruba I updated to use the `PackageBuilder` StringFormatConverter as `StringsConverter` seems functionality is in there and `SymplifyKernel'`s StringsConverter no longer exists in version `^10.0`

samsonasik:

packages/Demo/Entity/RectorRun.php
TomasVotruba: Yes, thank you. That was intended only for internal Kernel naming, not sure what is doing here.

packages/Demo/Entity/RectorRun.php
```diff
@@ -18,10 +18,10 @@
...
 
 #[Entity]
-class RectorRun implements TimestampableInterface
+final class RectorRun implements TimestampableInterface
```
TomasVotruba, line 24: This will probably fail in Doctrine.

samsonasik: Reverted https://github.com/rectorphp/getrector.org/commit/896cd5908a50fca9f269db38e6b80cd517e9b6ff

samsonasik:

rector.php
```diff
@@ -38,5 +39,11 @@
 
     $containerConfigurator->import(LevelSetList::UP_TO_PHP_80);
 
-    $parameters->set(Option::SKIP, ['*/var/cache/*', __DIR__ . '/packages/Demo/data/DemoFile.php']);
+    $parameters->set(Option::SKIP, [
+        '*/var/cache/*',
+        __DIR__ . '/packages/Demo/data/DemoFile.php',
```
samsonasik, line 44: @TomasVotruba the `DemoFile.php` is on root `./data' directory, do you think it can be removed from the skip option list as the file is not exists in the packages dir?

samsonasik: I removed it from the skip config list of ecs, phpstan, and rector.php as file moved to root `data` directory https://github.com/rectorphp/getrector.org/pull/441/commits/55dddf42c495b3106f29d98b5510c9f3390922c4

TomasVotruba: Ok :+1:

TomasVotruba (approved):

TomasVotruba:
Feel free to merge :+1:

composer.lock
```diff
@@ -3558,25 +3554,25 @@
         },
         {
             "name": "rector/rector",
-            "version": "0.11.60",
+            "version": "dev-main",
```
samsonasik, line 3557: by requiring phpstan ^1.0 with minimum-stability dev, it seems we will have `"rector/rector": "dev-main"` in the composer.lock before next rector release. The site is using php ^8.0, so I think it is fine so we can catch the bug early.

rector.php
samsonasik: @TomasVotruba it seems there is 500 error at demo page now after merge

Screen Shot 2021-11-07 at 18 27 37

I will revert it to see if that related to it.

TomasVotruba:
@samsonasik This whole PR should be reverted, before we figure out what happened.  Including dependencies
Now the demo page is down and people can't report issues or try Rector.

TomasVotruba:
The partial manaul revert of this PR https://github.com/rectorphp/getrector.org/commit/241490948f363803578b9a054c22e6fe17238e6b makes it hard to revert easily.

I'll have to revert last 2 commits with force push. Let's see if it helps

samsonasik:
@TomasVotruba ok, the demo page seems working ok now. Thank you 👍

TomasVotruba:
I see it too... https://getrector.org/demo/1ec3fd64-aea9-6504-9729-3dcbb9828cf0

Next time please revert the PR immediately to last working commit hash, if there is any 500 on the website. Then we can fix it locally in peace, instead of stressing with broken website.

Thanks :+1: 

IMO it will be related with Symfony 5.4 used here and Symfony 5.3 used in Rector. I had some simliar issues on rector/rector. We should wait for release of Symfony 5.4 stable and then use it here and in Rector itself.

PR #442, opened 2021-11-07
Revert remove packages/data/demo/DemoFile.php from the skip config list

samsonasik:
Current demo page got 500 error on "Process" button clicked, this revert unregister `packages/Demo/data/DemoFile.php` from skip config.

This reverts commit 55dddf42c495b3106f29d98b5510c9f3390922c4.

### DIFF
--- a/ecs.php
+++ b/ecs.php
@@ -3,6 +3,7 @@
 declare(strict_types=1);
 
 use PhpCsFixer\Fixer\ClassNotation\ClassAttributesSeparationFixer;
+use PhpCsFixer\Fixer\Strict\DeclareStrictTypesFixer;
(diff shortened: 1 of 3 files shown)


samsonasik:
@TomasVotruba the demo page seems still error even after revert it

Screen Shot 2021-11-07 at 18 27 37

Do you have idea for it?

samsonasik:
I tried locally, and it seems there is no error after click "Process" button at the demo page and data inserted to the table:

Screen Shot 2021-11-07 at 19 18 59

@TomasVotruba could you verify?

samsonasik:
The 500 error seems unrelated to skipping it as I tried both version with or without this skip and got inserted data into table locally.

@TomasVotruba could you verify?

PR #445, opened 2021-11-29
Upgrade to Rector 0.12, PHPStan 1.0, Symplify 10 and npm

TomasVotruba:
- bump deps
- add missing StingsConverter


TomasVotruba:
One more blocker

image

TomasVotruba:
@samsonasik Just FYI, I tried to update everything and deployed to the website. The demo works even 2 hours after deploy :partying_face: 

The main bug in previous PR was IMO this one:
https://github.com/rectorphp/getrector.org/pull/445#issuecomment-981681315

We're using `json` output format for Rector here, but the `SymfonyStyle` still outputed debug strings. That caused `json_decode(...)` to crash.

I fixed that here: https://github.com/rectorphp/rector-src/pull/1343, so the `SymfonyStyle` is always silent everywhere, if the output formatter is `json`.

That fixed the issue we had couple weeks ago.

PR #453, opened 2021-12-10
[Dep] Update rector version to ^0.12.7

samsonasik:
So only update when there is new version.


samsonasik:

TomasVotruba:
Thank you

PR #458, opened 2021-12-21
bump deps

TomasVotruba:


samsonasik:
@TomasVotruba it needs rebase and re-generate as the "remove sensio fw" commit reverted as cause demo page submit error https://github.com/rectorphp/getrector.org/issues/461

TomasVotruba:
Replaced by https://github.com/rectorphp/getrector.org/pull/472

PR #459, opened 2021-12-21
remove sensio fw

TomasVotruba:


samsonasik:
@TomasVotruba it seems cause error at demo page submit, ref https://github.com/rectorphp/getrector.org/issues/461 as downgrade back to rector 0.12.8 seems still error.

PR #463, opened 2021-12-22
Revert Remove sensio fw

samsonasik:
Reverts rectorphp/getrector.org#459

Per issue https://github.com/rectorphp/getrector.org/issues/461


samsonasik:
@TomasVotruba revert the remove sensio fw seems solve the demo page submit error 

https://getrector.org/demo/1ec62bf8-bf08-6e1c-85b1-ad1b9028bf0b

🎉

PR #464, opened 2021-12-22
Re-update back to rector 0.12.9

samsonasik:
@TomasVotruba let's try re-update back to rector 0.12.9 as the `remove sensio fw` was the reason the demo page was error https://github.com/rectorphp/getrector.org/issues/461#issuecomment-999191625

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -2617,21 +2617,21 @@
         },
         {
             "name": "rector/rector",
-            "version": "0.12.8",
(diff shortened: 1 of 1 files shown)


samsonasik:
@TomasVotruba it works 🎉  https://getrector.org/demo/1ec62c39-2eeb-6b5e-9b76-c55d5220bbde

PR #512, opened 2022-02-23
Update 2021-10-18-how-all-frameworks-can-bump-to-php-81-and-you-can-u…

ghispi:
…se-older-php.md

changed php version in the `rector-downgrade.php` to keep it with description provided above


TomasVotruba:
Thank you :+1:

PR #533, opened 2022-04-15
Update to use latest PHPStan ^1.5.6

samsonasik:
as composer.json using `rector/rector` `dev-main`, PHPStan need to use PHPStan ^1.5.6 in composer.lock and composer.json due to https://github.com/rectorphp/rector-src/pull/2014


TomasVotruba:
👍

PR #536, opened 2022-04-18
Fix First Run link linked when open from non homepage

samsonasik:
When user open different page, eg: `/blog`, then click `First Run` link, it currently only append `#first-run` instead of go to home page with `#first-run` id: `https://getrector.org/blog#first-run`. 

This PR fix it.


TomasVotruba:
Thanks, I noticed it yesterday and was too lazy to fix :laughing:

PR #549, opened 2022-04-27
fix 404 twitter link

rajyan:
I've changed my twitter account id (I didn't know that the redirect is not permanent...). 
I would appreciate it if you could deploy this when you have time! I'm not in a hurry with this fix.


samsonasik:
I set setting website to be deployed on "At 00:05 on Sunday, Wednesday, and Friday." https://crontab.guru/#5_0_*_*_0,3,5 so it should be deployed in Friday.

samsonasik:
Thank you @rajyan

rajyan:
Thank you too!

PR #613, opened 2022-06-24
[GithubMagicLink] Handle Create a Test link on rector-* packages

samsonasik:
Based on https://github.com/rectorphp/rector-symfony/issues/188#issuecomment-1165463915 to make Create a Test link pointed to rector-* packages when use rule from them.


samsonasik:
implemented 🎉

TomasVotruba (approved):

PR #631, opened 2022-07-05
Fix amnesia on routes.php config error

samsonasik:
**Before**

Screen Shot 2022-07-05 at 17 27 39

**After**

Screen Shot 2022-07-05 at 17 30 00

Fixes https://github.com/rectorphp/getrector.org/issues/630

### DIFF
--- a/config/routes.php
+++ b/config/routes.php
@@ -3,10 +3,9 @@
 declare(strict_types=1);
 
 use Symfony\Component\Routing\Loader\Configurator\RoutingConfigurator;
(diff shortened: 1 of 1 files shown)


samsonasik:
Let's merge it to avoid next hour scheduled build error on getrector.org at 12:05

TomasVotruba:
Thanks :+1:

TomasVotruba:
I've added it to PHPStan so we have it covered next time:
https://github.com/rectorphp/getrector.org/commit/38aed13804658787191141a0243bee956166c09c

PR #707, opened 2022-08-14
[GitHubMagicLink] Remove Nette from FixtureLinkFactory as the rector-nette package removed

samsonasik:


samsonasik:

PR #708, opened 2022-08-14
[GitHubMagicLink] Update rector-downgrade-php link create fixture to use rules-tests instead of tests

samsonasik:
@TomasVotruba due to your PR:

- https://github.com/rectorphp/rector-downgrade-php/pull/15

The fixture create link needs to be updated from `tests` to `rules-tests`


samsonasik:

PR #718, opened 2022-08-20
Update back rector to use dev-main

samsonasik:
@TomasVotruba your commit https://github.com/rectorphp/getrector.org/commit/58f6f0e60df141909df605b0702fbf1c7cd4f4ab was update to use rector ^0.14, that will make per-6 hours update not applied as will wait for next release.

This PR update back to use dev-main to get faster feedback after fix applied to dev-main.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
         "ext-json": "*",
         "erusev/parsedown-extra": "^0.8.1",
         "jajo/jsondb": "^3.0",
-        "rector/rector": "^0.14",
+        "rector/rector": "dev-main",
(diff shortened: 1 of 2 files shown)


samsonasik:

TomasVotruba:
I see. My goal was to trigger update to the latest release.
I wonder we could show the current commit + tag version on Rector demo page. It would make reports clear to specific version.

samsonasik:
Maybe we can use https://packagist.org/packages/ocramius/package-versions

samsonasik:
https://github.com/rectorphp/getrector.org/pull/719

PR #719, opened 2022-08-20
Display Rector version at demo page

samsonasik:
@TomasVotruba this is display Rector vesion at demo page:

Screen Shot 2022-08-20 at 17 10 58

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,6 +7,7 @@
         "ext-json": "*",
         "erusev/parsedown-extra": "^0.8.1",
         "jajo/jsondb": "^3.0",
+        "ocramius/package-versions": "^2.5",
         "rector/rector": "dev-main",
         "symfony/asset": "^6.1",
         "symfony/cache": "^6.1",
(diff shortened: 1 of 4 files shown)


samsonasik:

samsonasik:
rebased.

TomasVotruba:
Thank you :+1:

TomasVotruba:
Btw, could you make the text grey, so it does not take that much attention?

The day of version release could be also important, to determine how old the deployed version is. Could you add it?

samsonasik:
I will try, the PackageVersion doesn't show the date, so it probably need to directly read from composer.lock

https://github.com/rectorphp/getrector.org/blob/872a6ae0fd1a99c41fd2dea288a8f0860c57ba7d/composer.lock#L667

TomasVotruba:
The date time is stored in this constant:
https://github.com/rectorphp/rector/blob/0be9fe255157e4140c86c4dbbfa9a347f3ae6997/src/Application/VersionResolver.php#L25

samsonasik:
https://github.com/rectorphp/getrector.org/pull/721

PR #813, opened 2022-10-13
Add airy design with business focus + add documentation + bump to Twitter Bootstrap 5, Font Awesome 6

TomasVotruba:


.env_acceptance.dist
```diff
@@ -1,7 +0,0 @@
-###> symfony/framework-bundle ###
-APP_ENV=prod
-APP_DEBUG=0
-APP_SECRET=67d829bf61dc5f87a71fd814e2c9f629
-###< symfony/framework-bundle ###
-
-DEBUG_COOKIE=
```
samsonasik, line 7: This file is needed for acceptance test show production env https://github.com/rectorphp/getrector.org/blob/558aca6b5b71a62d38955cc1eb5ef48fe1e30fdd/.github/workflows/acceptance_test.yaml#L25

package.json
```diff
@@ -1,12 +1,13 @@
...
     "name": "getrector-org",
     "dependencies": {
-        "@fortawesome/fontawesome-free": "^5.9.0",
-        "@fortawesome/free-brands-svg-icons": "^5.9.0",
-        "@fortawesome/free-regular-svg-icons": "^5.9.0",
-        "@fortawesome/free-solid-svg-icons": "^5.9.0",
+        "@fortawesome/fontawesome-free": "^6.2",
+        "@fortawesome/free-brands-svg-icons": "^6.2",
+        "@fortawesome/free-regular-svg-icons": "^6.2",
+        "@fortawesome/free-solid-svg-icons": "^6.2",
```
samsonasik, line 7: Current deploy script on getrector.org doesn't run 

```
node install && yarn dev
```

so probably require to update after deployed

samsonasik:
Ref for why acceptance test is needed https://github.com/rectorphp/getrector.org/pull/723

.env_acceptance.dist
TomasVotruba: Fixed

package.json
TomasVotruba: Could you add that line to the deploy script now? 

So we don't have to think about it in the future and have it covered :+1:

samsonasik: I can see `yarn build` in the script, is that same with `node install && yarn dev` ?

TomasVotruba: I'm not much expert on this one :) 

Locally I've run:

```bash
yarn upgrade # similar to "composer update"
yarn dev # to build the assets
```

samsonasik: It seems `npm install` actually needed, to download the updated packages first, I updated the script so it install the dependencies and build the webpack

samsonasik: me too, I currently apply: `npm install && yarn build`, yarn build seems for production so it minified result on generated webpack, let's see if that's working on next build

TomasVotruba:
## How does the changes look like? :) 

1_demo_before
2_demo_after
3_home_before
4_home_after

TomasVotruba:
Let's :ship: this, so we can iterate on it :)

TomasVotruba:
It's on: https://getrector.org/ :partying_face:

PR #896, opened 2022-11-29
Remove TypedPropertyRector from config DEMO page

samsonasik:
Since TypedPropertyRector is now deprecated:

https://github.com/rectorphp/rector-src/blob/2b35d2e75ea03ffe643c012485952ec6df3ea2ba/rules/Php74/Rector/Property/TypedPropertyRector.php#L78-L81

The Demo page should not add it to the config example, I change it to `TypedPropertyFromAssignsRector`


samsonasik:

TomasVotruba:
Thanks :+1:

PR #902, opened 2022-11-30
Add docs + add tests

TomasVotruba:
- headline
- misc
- add docs section


data/docs/how-to-add-test-case.md
```diff
@@ -0,0 +1,146 @@
...
+
+`FinalizeClassesWithoutChildrenRector`
+
+↓
+
+`FinalizeClassesWithoutChildrenRectorTest` (test class)
+
+Right here:
+
+![Rule Test Case](/docs/images/docs_rule_test_case.png)
```
samsonasik, line 70: images seems need to be imported, or use raw uri?  https://github.com/rectorphp/rector/blob/main/docs/images/docs_applied_rules.png

TomasVotruba: I've missed to move it here, thanks :+1:

TomasVotruba: See https://github.com/rectorphp/getrector.org/pull/904

PR #913, opened 2022-12-03
Fix docs rector rules list URL

samsonasik:
Currently got 404 due to relative path that goes to https://getrector.org/docs/rector_rules_overview.md


TomasVotruba:
Thanks

PR #981, opened 2023-01-06
Fix cannot apply patch generic-php-config-loader.patch

samsonasik:
Running composer install/update seems currently got error:

```bash
  - Applying patches for symfony/dependency-injection
    https://raw.githubusercontent.com/symplify/vendor-patch-files/main/patches/generic-php-config-loader.patch (0)
   Could not apply patch! Skipping. The error was: Cannot apply patch https://raw.githubusercontent.com/symplify/vendor-patch-files/main/patches/generic-php-config-loader.patch
```

it seems due to current requirement of symplify/coding-standard can't apply the symfony/dependency-injection 6.2+.

This PR fix it by downgrade the symfony/dependency-injection to 6.1.*

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
         "rector/rector": "dev-main",
         "symfony/asset": "^6.1",
         "symfony/cache": "^6.1",
-        "symfony/dependency-injection": "^6.1",
+        "symfony/dependency-injection": "6.1.*",
         "symfony/dotenv": "^6.1",
         "symfony/form": "^6.1",
         "symfony/framework-bundle": "^6.1",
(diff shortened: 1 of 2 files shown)


samsonasik:

samsonasik:
Looking at composer.lock, it seems due to symplify/coding-standard

https://github.com/symplify/coding-standard/blob/d7d5bc24c52ba766518641e847cad95151ebcfb7/composer.json#L42

@TomasVotruba should the symplify/coding-standard require symfony di 6.1.* ?

TomasVotruba:
> should the symplify/coding-standard require symfony di 6.1.* ?

:+1: Yes, as the downgrade of 6.2 does not work for autowiring propperly. Neither does the patch.

PR #1007, opened 2023-01-16
Fix final config suggestion

DannyvdSluijs:
This fixes the suggested config by adding the missing opening tag and the missing closing bracket and semi-colon. Now if the example is copied and pasted into a `rector.php` file it works straight out of the box.


TomasVotruba:
Thank you :+1:

PR #1043, opened 2023-02-01
Use correct method name

reneroboter:
I can't find the `reconstruct ` method in the rector source code. After checking the `Custom Rule` page I guess the `refactor` method is meant. 

If I'm wrong some explanation would be fine :)


TomasVotruba:
Thank you, this is really from prehistorical times :D

PR #1046, opened 2023-02-02
Remove nette/utils dependency in custom-rule.md

### LINKED ISSUE #1044: nette/utils is required for the custom rule example
I created my first own Rector and by the way I realised that the original example depend on `nette/utils`. I guess when someone unexpired will try this example he will be confused, because after installing rector as dev dependency you don't have `nette/utils`. 

So my recommendation would be to get rid of `nette/utils` or add a hint that you've to install `nette/utils` as dependency.

https://github.com/rectorphp/getrector.org/blame/main/data/docs/custom-rule.md#L19

reneroboter:
@TomasVotruba this should fix #1044 :)

### DIFF
--- a/data/docs/custom-rule.md
+++ b/data/docs/custom-rule.md
@@ -16,7 +16,6 @@ declare(strict_types=1);
 
 namespace Utils\Rector\Rector;
 
-use Nette\Utils\Strings;
 use PhpParser\Node;
 use PhpParser\Node\Identifier;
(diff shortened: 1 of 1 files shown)


samsonasik (approved):

samsonasik:
Thank you @reneroboter

TomasVotruba:
Thank you :clap:

PR #1073, opened 2023-02-14
Add documentation on how to write tests for custom rules.

ian-zunderdorp:
I noticed while I was creating a custom rule that there was no good documentation on how to write tests, only a recommended file structure without clarification. There is an implicit expectation that tests are simply copied from existing rules.

It is debatable whether this needs to be documented explicitly, because documentation can get out of date, or whether the documentation for recommended file structure should explicitly mention the expectation of copying from existing rules. If you prefer to not document writing tests for custom rules, let me know and I'll create a PR for making the expectation explicit.

While describing `provideData`, I wanted to link to another page (How to add test case) as the structure and logic of the Fixture files was already described elsewhere. However, there is no prior example of this and I'm unsure what the preferred approach is. Possible options for linking are:
- Full absolute URL to the .md file on GitHub
  - Advantage: Will always work as long as repository stays at same location
  - Downside: getrector.com documentation redirects to GitHub. Also makes it harder to provide correctly linking versioned documentation
- Full absolute URL to the rendered documentation at https://getrector.com/documentation/
  - Advantage: Will always work, provides the "nicer" documentation interface
  - Downside: URL has to be hardcoded, making it harder to provide correctly linking versioned documentation later or to migrate to a different URL / subroot
- Relative URL to the .md file
  - Advantage: Easier to implement, supports migrating repository and versioned documentation, better support for local development
  - Downside: Broken link on getrector.com
  - Possible improvement to fix downside: Implement extra logic in `HTMLFromMarkdownFactory` to convert relative URLs to .md files
- Relative URL to the rendered documentation
  - Advantage: Still easy to implement, supports migrating to different URL / subroots and versioned documentation if rendered documentation is published as versions
  - Downside: Broken link in GitHub
- Keep as-is: Least effort but worst developer experience

### DIFF
--- a/data/docs/custom-rule.md
+++ b/data/docs/custom-rule.md
@@ -114,6 +114,56 @@ rector.php
 composer.json
 ```
 
+### Writing tests
+
+Rector provides a structured way of running your rules on snippets of code so you can validate that your rule works as expected.
+It uses PHPUnit to run the tests and `rector/rector` provides the `AbstractRectorTestCase` class to simplify test configuration.
+
+The usual structure of the test class is as follows:
+```php
+declare(strict_types=1);
+
+namespace Utils\Rector\Tests\Rector\MyFirstRector;
+
+use Iterator;
+use PHPUnit\Framework\Attributes\DataProvider;
+use Rector\Testing\PHPUnit\AbstractRectorTestCase;
+
+final class MyFirstRectorTest extends AbstractRectorTestCase
+{
+    #[DataProvider('provideData')]
+    public function test(string $filePath): void
+    {
+        $this->doTestFile($filePath);
+    }
+
+    public static function provideData(): Iterator
+    {
+        return self::yieldFilesFromDirectory(__DIR__ . '/Fixture');
+    }
+
+    public function provideConfigFilePath(): string
+    {
+        return __DIR__ . '/config/configured_rule.php';
+    }
+}
+```
+
+You can see that there are 3 functions in this test class:
+
+- `public function test(string $filePath): void`:
+  - This method is to help PHPUnit detect this test
+  - For `$filePath`, we use a [PHPUnit DataProvider](https://phpunit.readthedocs.io/en/10.0/writing-tests-for-phpunit.html#data-providers)
+  - This triggers a run for every test file in your Fixtures directory
+- `public static function provideData(): Iterator`:
+  - As stated above, this is a PHPUnit DataProvider
+  - Using `self::yieldFilesFromDirectory` it iterates over all test cases you provided
+  - See `How To Add Test Case` for the files that are expected in the `/Fixture` directory
+- `public function provideConfigFilePath(): Iterator`:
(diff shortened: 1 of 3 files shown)


ian-zunderdorp:
Hi @TomasVotruba,

What is the best way to go forward with this pull request? Is this even something you want?

TomasVotruba:
Hi @ian-zunderdorp ,

thanks for the ping. Let's keep this going.

Could you add it as a new item? Writing custom rules and testing them on one page might be too much to handle for someone who is starting.

<br>

Could you add an example of very simple test fixture? 

* To show there should be `-----` between before and after code
* The suffix must be `*.php.inc` to avoid picking up by automated tools

<br>

Last step, it could be useful to see the actual directory/file structure with config, rule and tests, something like:

```bash
/tests
    /Rector
         /ClassMethod
               RenameClassMethodRector.php
               /config
                   configured_rule.php
               /Fixtures
                   some_file.php.inc
````

<br>

Thanks

ian-zunderdorp:
Question: Is there a specific flow that you expect for pull requests? If you want, I could create new pull requests to add/update CONTRIBUTING.md where needed, so next time people contribute for the first time they know what to expect and how to get pull requests approved. By default I don't mention specific maintainers unless it is explicitly listed in CONTRIBUTING.md

TomasVotruba:
There is nothing I'm able to define, just to make CI pass and content clear.

ian-zunderdorp:
> Could you add it as a new item? Writing custom rules and testing them on one page might be too much to handle for someone who is starting.

I'll start working on this.

> Could you add an example of very simple test fixture?
> 
> * To show there should be `-----` between before and after code
> * The suffix must be `*.php.inc` to avoid picking up by automated tools
> 
> Last step, it could be useful to see the actual directory/file structure with config, rule and tests, something like:
> 
> ```shell
> /tests
>     /Rector
>          /ClassMethod
>                RenameClassMethodRector.php
>                /config
>                    configured_rule.php
>                /Fixtures
>                    some_file.php.inc
> ```

I can do this (and I'll also include an example for skipped tests). The main reason I did not do this yet was because the "How To Add Test Case" already contains an example of writing simple test fixtures & the "Write custom rule" page already contains an example directory structure. Duplicating this introduces a higher maintenance burden as there are now two pages on the same topic, that is why I was considering adding hyperlinks.

In the meantime I'll start working on moving everything to a new page.

> There is nothing I'm able to define, just to make CI pass and content clear.

:+1:

TomasVotruba:
> Duplicating this introduces a higher maintenance burden as there are now two pages on the same topic, that is why I was considering adding hyperlinks.

I see. The "How to add test case" is rather about reporting an issue to Rector project itself. The title could be renamed to better show this.

ian-zunderdorp:
@TomasVotruba I've moved my changes to a new page, including examples as mentioned before. I haven't applied any changes yet to the "How to add test case", do you want a new pull request for that or include it here?

TomasVotruba:
> do you want a new pull request for that or include it here?

New PR would be better, to keep changes isolated.

Thanks for the quick work, I'll merge it  :+1:

PR #1082, opened 2023-02-20
docs: make example easier to copy/paste

staabm:


TomasVotruba:
This might turn all the long classes to scroll bar mess.

Instead try Enhanced GitHub plugin. It adds copy paste button to all the code snippets.

staabm:
my main idea was more related to: make code copy pastable in a single step.

as is - when integrating the snippet into a pre-existing `rector.php` - I need copy/paste 2 times. once for the `use ..` and once for the actual code lines.

TomasVotruba:
I got that, but it would turn docs into mess of randoms classes that some are copy-pasteable and some now.
We use the PSR-12 standard here.

This should be handled on PHPStorm side, I know it worked before very well.

PR #1089, opened 2023-02-22
Rename the "How to add test case" page

ian-zunderdorp:
As discussed in #1073 

Also added a leading statement to explain why a test case in a pull request is needed.


ian-zunderdorp:
@TomasVotruba There you go

TomasVotruba:
Thanks, very well done :clap:

PR #1150, opened 2023-03-06
[post] Add setup-ci command

TomasVotruba:
- kick of post
- styling
- misc


composer.json
```diff
@@ -8,7 +8,7 @@
         "imagine/imagine": "^1.3",
         "jajo/jsondb": "^3.0",
         "laravel/framework": "^10.1",
-        "rector/rector": "dev-main",
+        "rector/rector": "^0.15.20",
```
samsonasik, line 11: dev-main should be used for faster feedback

TomasVotruba: I'm testing here the tagged version, as this post is about a freshly tagged feature :+1:

PR #1393, opened 2023-06-08
[automated] Update Rector Dependencies

TomasVotruba:
Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action

PR #1394, opened 2023-06-10
Fix unit test of removed RemovedAndAddedFilesCollector

samsonasik:
Ref https://github.com/rectorphp/getrector-com/pull/1393#issuecomment-1585758237


samsonasik:
All checks have passed 🎉 @TomasVotruba let's merge it ;)

PR #1437, opened 2023-06-24
[automated] Update Rector Dependencies

TomasVotruba:
Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action


samsonasik:
@TomasVotruba changing symfony rules under `rules` directory make link no longer valid to point to correct directory fixture.

The `FixtureLinkFactory` need to be refactored.

samsonasik:
Not every rules is under `rules` for rector symfony, 

https://github.com/rectorphp/rector-symfony/tree/main/src/Rector/ClassMethod

and it also has different prefix, eg: `CodeQuality` instead of `SymfonyXy` so it need dynamic check 

https://github.com/rectorphp/rector-symfony/tree/main/rules-tests/CodeQuality/Rector

samsonasik:
@TomasVotruba fixed ;)

PR #1496, opened 2023-07-17
Bump dependencies

samsonasik:
To see if `Call to undefined method Rector\\Core\\PhpParser\\Printer\\BetterStandardPrinter::pPHPStan_Node_AlwaysRememberedExpr` error still reproduced with latest deps

Ref https://github.com/rectorphp/rector/issues/8063#issuecomment-1637604533
Ref https://getrector.com/demo/bf537a77-1c66-474c-b415-9b2794835341
Ref


samsonasik:
It seems still reproduced https://getrector.com/demo/8c6e600c-6537-4895-b945-ab38e9db9357 , not sure why it can't be reproduced via unit test and e2e test in rector-src

PR #1636, opened 2023-09-06
Use SetList::TYPE_DECLARATION in demo defaults

staabm:
I think for an end-user its more suprising and compelling to see rector inferring types instead of detecting/removing dead code

### DIFF
--- a/resources/demo/DemoFile.php
+++ b/resources/demo/DemoFile.php
@@ -2,11 +2,17 @@
 
 final class DemoFile
 {
-    public function run()
+    public function run(bool $param)
     {
-        return 5;
+        if ($this->isTrue($param)) {
+            return 5;
+        }
 
-        // we never get here
-        return 10;
+        return '10';
+    }
+
+    private function isTrue($value)
+    {
+        return $value === true;
     }
 }
--- a/resources/demo/demo-config.php
+++ b/resources/demo/demo-config.php
@@ -7,7 +7,7 @@ use Rector\TypeDeclaration\Rector\Property\TypedPropertyFromAssignsRector;
 return static function (RectorConfig $rectorConfig): void {
     // A. run whole set
     $rectorConfig->sets([
-        SetList::DEAD_CODE,
+        SetList::TYPE_DECLARATION,
     ]);
 
     // B. or single rule


TomasVotruba:
Hi, thanks for the proposal. Makes sense to me to show more powerful contents :+1: 

Could you share screenshot what is the submitted file result?

staabm:
the diff will be

grafik

I was not able to run the rector demo on my machine.
it errors with "php command not found" errors. not sure why.

TomasVotruba:
We should fix that first, so you can develop with real feedback.

The validation is triggered here: https://github.com/rectorphp/getrector-com/blob/821e48e3275bd02e327648e59f2e217c994d15ea/src/Validator/PhpLinter.php#L29-L30

It should be bare php-parser run. Where exactly is this error thrown?

samsonasik:
@TomasVotruba php command not found is unrelated with this change, it mostly path issue setting local dev machine ( mostly due to multiple php installed and composer/symfony/daimon can't know which php installed)

staabm:
@samsonasik can you give me a hint where this is configured, or how I can make the getrector-com project make use the correct `php` command?

samsonasik:
@staabm I don't know, I have that issue as well locally 😄 , all I can do is pray it not break something for testing demo page 😁

staabm:
the php-command-not-found problem is gone, but new errors are coming up. 

I was able to work arround the "failed to create temp dir error"

266045624-0611bf51-f8fa-4efc-81ef-1c8402a56e34

using

```diff
diff --git a/src/DemoRunner.php b/src/DemoRunner.php
index cb78094..02a64a3 100644
--- a/src/DemoRunner.php
+++ b/src/DemoRunner.php
@@ -78,6 +78,8 @@ final class DemoRunner

         $process = new Process([
             PHP_BINARY ,
+            '-d',
+            'sys_temp_dir='.sys_get_temp_dir(),
             // paths for phpunit differs based on test/demo, not sure why
             \defined('PHPUNIT_COMPOSER_INSTALL') ? 'vendor/bin/rector' : '../vendor/bin/rector',
             'process',
```

but now I get tcp errors

grafik

samsonasik:
@staabm that's strange, I cannot reproduce. You may need to manually create rector_cached_files dir in C:\Windows

staabm:
could someone of you guys test this PR and merge it if successfull. I guess the app is not working on my corporate laptop for some security/safety tooling reason

samsonasik:
Tested locally, it seems working ok, thank you @staabm

samsonasik (approved):

staabm:
Thanks!

TomasVotruba:
:+1:

PR #1924, opened 2024-01-11
Upgrade to Laravel 11 🎉🎉🎉

TomasVotruba:


samsonasik:
I will need bump server to php 8.2 first, I will do today.

samsonasik:
I tried upgrade to php 8.2 on server, it currently got blank page on execute demo page, looking at apache error log, it got memory exhausted, looking at `phpinfo()`, it reads as `128MB`, I changed various php.ini and restart apache and it doesn't take effect. 

I will look more...

samsonasik:
I see, it seems need to be forced restart via fpm:

```
/etc/init.d/php8.2-fpm restart
```

not only apache. I upgraded server to php 8.2 and it seems working ok 🎉 

https://getrector.com/

composer.json
```diff
@@ -61,18 +59,13 @@
...
         "fix-cs": "vendor/bin/ecs check --fix --ansi",
-        "phpstan": "php -d memory_limit=1G vendor/bin/phpstan analyse --error-format symplify --ansi",
+        "phpstan": "vendor/bin/phpstan analyse --ansi --error-format symplify",
         "rector": "vendor/bin/rector process --ansi"
     },
     "config": {
         "sort-packages": true,
         "allow-plugins": {
-            "phpstan/extension-installer": true,
-            "cweagans/composer-patches": true,
```
samsonasik, line 71: running composer install locally got notice:

```
➜  getrector.org git:(tv-laravel-11) composer install
cweagans/composer-patches contains a Composer plugin which is currently not in your allow-plugins config. See https://getcomposer.org/allow-plugins
Do you trust "cweagans/composer-patches" to execute code and wish to enable it now? (writes "allow-plugins" to composer.json) [y,n,d,?] n
```

I guess it can be keep here to avoid it composer install not executed on server while files already added.

samsonasik:
I tried locally, it got error:

```
Unsupported cipher or incorrect key length. Supported ciphers are: aes-128-cbc, aes-256-cbc, aes-128-gcm, aes-256-gcm.
```

Screenshot 2024-01-11 at 12 39 06

Looking at https://tecadmin.net/resolving-unsupported-cipher-or-incorrect-key-length-error-in-laravel/ , the config key need to be re-generated:

```
php82 artisan key:generate
```

samsonasik:
I tested at demo page execute, and it seems now got `Page Expired` message:

Screenshot 2024-01-11 at 12 52 26

samsonasik:
So, it seems csrf fields needs to be added to `resources/views/demo/demo.blade.php`, adding csrf field like this make it works:

```diff
        <form action="{{ action(\Rector\Website\Http\Controllers\ProcessDemoFormController::class) }}"
              method="post">
            
+            @csrf <!-- {{ csrf_field() }} -->
            
             @if ($rector_run->hasRun() && $rector_run->isSuccessful() !== true)
```

TomasVotruba:
Thanks for battle testing 🙂👌

Could you push here your commits and update `firsr-run.sh` with all the required steps?
Thanks

composer.json
samsonasik: Remove vendor and run composer update seems solve it. I will verify once more once this PR deployed.

samsonasik:
`firsrt-run.sh` seems fine 👍

TomasVotruba:
Thank you :+1:  I'll try it locally and if all is good, merge it

TomasVotruba:
All good locally. Let's ship  it :rocket:

database/migrations/2024_01_11_014957_create_cache_table.php
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * Reverse the migrations.
+     */
+    public function down(): void
+    {
+        Schema::dropIfExists('cache');
+        Schema::dropIfExists('cache_locks');
+    }
+};
```
TomasVotruba, line 35: @samsonasik I wonder if we can drop there, but it seems `clear:cache` requires a database for some reason :thinking:

PR #1930, opened 2024-01-13
Slightly improved the clarity of the parallel docs

BelleNottelling:
Hi all,
Thanks for making Rector, it's proved to be a very handy tool!

This pull request just makes a small change to the "Troubleshooting Parallel" documentation to improve the clarity on how you actually set these options.

### DIFF
--- a/resources/docs/troubleshooting-parallel.md
+++ b/resources/docs/troubleshooting-parallel.md
@@ -25,8 +25,8 @@ s you have to process rector.
 There are 3 options we can define for parallel processing:
 ```php
 public function parallel(
-    int $seconds = 120,
-    int $maxNumberOfProcess = 16,
+    int $seconds = 60,
+    int $maxNumberOfProcess = 32,
     int $jobSize = 20
 ): void
 ```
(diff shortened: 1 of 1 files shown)


resources/docs/troubleshooting-parallel.md
```diff
@@ -32,6 +31,11 @@ public function parallel(
 ): void
 ```
 
+To manually set these options, call the `parallel` function within the Rector config class:
+```php
+$rectorConfig->parallel(120, 16, 20);
```
samsonasik, line 36: Rector 0.19.0 is using `60, 32, 20` as default value, ref

https://github.com/rectorphp/rector-src/blob/ac072bc8af186b2c86256aa7d29e52bb608ece27/src/Config/RectorConfig.php#L91

Could you update the default value on line 28 to 30 above it?

BelleNottelling: Done :)

samsonasik:
Thank you @BelleNottelling

BelleNottelling:
> Thank you @BelleNottelling

You're welcome!
Thanks again for Rector, it really is a great tool

PR #1962, opened 2024-01-22
Fix error 500

samsonasik:


config/database.php
```diff
@@ -0,0 +1,5 @@
+<?php
+
+return [
+    'DB_DATABASE' => __DIR__ . '/../database/database.sqlite',
```
samsonasik, line 4: @TomasVotruba this is needed due to now require absoute path.

.env.dist
```diff
@@ -5,4 +5,3 @@ APP_KEY=base64:c9yprqAy56EwJ1pvPeyRkP6xV3MTtsa498Hsffd3/rw=
 APP_URL=http://localhost
 
 DB_CONNECTION=sqlite
-DB_DATABASE=database/database.sqlite
```
samsonasik, line 8: the relative path need to updated to use absolute path on config/database.php.

samsonasik:
Ref https://github.com/rectorphp/rector/issues/8422

still error, seems some permission issue, I am still looking into it.

samsonasik:
It fixed now 👍

config/database.php
samsonasik: It seems only define database type on .env is enough, and will use database/database.sqlite, see:

- https://github.com/rectorphp/getrector-com/pull/1963

TomasVotruba:
Hi Abdul, thanks for quick fix. Do you know what it was caused by?
I think I didn't lock Laravel to specific commit, and there was a BC break.

samsonasik:
It was a php-fpm down, which result 503, restarting it got no path cache, updating laravel got 500, it was rough guess

PR #1998, opened 2024-02-01
Use withImportNames() method

gitomato:
The new `RectorConfig::configure()` method requires `withImportNames() ` method instead of `importNames()`.


TomasVotruba:
Thank you :pray:

Issue #2022, opened 2024-02-08
GitHub Actions Cache key


SomeBdyElse:
The current documentation suggests to use the following cache key for GitHub actions: 

```
${{ runner.os }}-rector-${{ hashFiles('**/composer.lock') }}
```

https://github.com/rectorphp/getrector-com/blob/be17729dedb91d48dbe6a904ba3f84bed391d28d/resources/docs/cache-in-ci.md?plain=1#L42

GHA caches are immutable. Therefore the cache will be stored once and updated only if the composer.lock changes. I do believe that the cache contents are dependent on non-vendor php files which might changes without alterations to the composer.lock. Therefore the cache should also be updated without changes to the composer.lock. 

I suggest to change the cache key to: 
```
rector-${{ hashFiles('**/composer.lock') }}-${{ github.run_id }}
```

The pattern is suggested here: https://github.com/actions/cache/blob/main/tips-and-workarounds.md#update-a-cache

staabm:
we usually use something like

```
key: ${{ steps.setup-php.outputs.php-version }}-rector-${{ github.run_id }}
restore-keys: ${{ steps.setup-php.outputs.php-version }}-rector- 
```

`${{ hashFiles('**/composer.lock') }}` is not really useful, I agree.

samsonasik:
I am not familiar with it, could you guys try create a PR for it? Thank you.

staabm:
just sent https://github.com/rectorphp/getrector-com/pull/2023

@SomeBdyElse could you double check it?

PR #2023, opened 2024-02-08
Improve GitHubActions default cache-keys

staabm:
closes https://github.com/rectorphp/getrector-com/issues/2022

### DIFF
--- a/resources/docs/cache-in-ci.md
+++ b/resources/docs/cache-in-ci.md
@@ -39,8 +39,8 @@ On GitHub Actions, you can use the [built-in cache action](https://github.com/ac
         uses: actions/cache@v3
         with:
           path: ./var/cache/rector
-          key: ${{ runner.os }}-rector-${{ hashFiles('**/composer.lock') }}
-          restore-keys: ${{ runner.os }}-rector-
+          key: ${{ steps.setup-php.outputs.php-version }}-rector-${{ github.run_id }}
+          restore-keys: ${{ steps.setup-php.outputs.php-version }}-rector- 
 
       - name: Rector Dry Run
         run: php vendor/bin/rector process --dry-run --config=rector.php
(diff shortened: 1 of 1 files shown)


resources/docs/cache-in-ci.md
```diff
@@ -39,8 +39,8 @@ On GitHub Actions, you can use the [built-in cache action](https://github.com/ac
         uses: actions/cache@v3
         with:
           path: ./var/cache/rector
-          key: ${{ runner.os }}-rector-${{ hashFiles('**/composer.lock') }}
-          restore-keys: ${{ runner.os }}-rector-
+          key: ${{ steps.setup-php.outputs.php-version }}-rector-${{ github.run_id }}
```
SomeBdyElse, line 42:
```suggestion
          key: ${{ runner.os }}-rector-${{ hashFiles('**/composer.lock') }}-${{ github.run_id }}
```

Hi @staabm !

* `${{ runner.os }}` still seems necessary, unless we set `enableCrossOsArchive`
* `${{ steps.setup-php.outputs.php-version }}` I do not have this step or output in my setups. So it would be confusing if the docs just assume its existence. 
* **/composer.lock seems neccessary to me, as it hashes the state of vendor php classes as well as the rector version.

SomeBdyElse (changes requested):

resources/docs/cache-in-ci.md
samsonasik: is that matter committed or not the composer.lock?

SomeBdyElse: The `hashFiles('**/composer.lock')` only makes sense if the `composer.lock` is committed. If `composer.lock` does not exist, the hash function will return an empty string (no harm done).

samsonasik: @staabm could you verify it and then we can merge it ?

staabm: its a matter of taste. IMO `setup-php` is used in any project I know which uses github action.
we don't use `${{ hashFiles('**/composer.lock') }}` in our projects and did not have problems .. therefore I would not use it and keep it simpler

staabm: no matter how we turn the example mentioned in the docs, its pretty unlikely that in a project it can be used 1:1

samsonasik:
Let's merge it, thank you guys 👍

PR #2030, opened 2024-02-09
configure and create cacheDirectory

staabm:
as discussed in https://github.com/rectorphp/rector/discussions/8357#discussioncomment-8420673

### DIFF
--- a/resources/docs/cache-in-ci.md
+++ b/resources/docs/cache-in-ci.md
@@ -26,7 +26,7 @@ Generate the cache on your development machine, by running the command:
 vendor/bin/rector process --dry-run
 ```
 
-You can find it in your repository directory under `./var/cache/rector/`, containing folders like `0a`, `0b`, `0c`, ... containing the cache objects representing the latest run.
+You can find it in `/tmp/rector`, containing folders like `0a`, `0b`, `0c`, ... containing the cache objects representing the latest run.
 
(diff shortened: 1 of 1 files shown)


resources/docs/cache-in-ci.md
```diff
@@ -12,7 +12,7 @@ return RectorConfig::configure()
         cacheClass: FileCacheStorage::class,
 
         // specify a path that works locally as well as on CI job runners
-        cacheDirectory: './var/cache/rector'
+        cacheDirectory: '/tmp/rector'
```
staabm, line 15: using `/tmp/rector` as it works well for use.

phpstan by default uses `/tmp/phpstan`

resources/docs/cache-in-ci.md
```diff
@@ -38,10 +38,12 @@ On GitHub Actions, you can use the [built-in cache action](https://github.com/ac
       - name: Rector Cache
         uses: actions/cache@v3
         with:
-          path: ./var/cache/rector
+          path: /tmp/rector
           key: ${{ runner.os }}-rector-${{ github.run_id }}
           restore-keys: ${{ runner.os }}-rector-
 
+      - run: mkdir -p /tmp/rector
```
samsonasik, line 45: should it run before line 38 above executed?

staabm: Nope. Needs to be here.

samsonasik:
Ok, thank you @staabm

PR #2034, opened 2024-02-10
[Doc] Add missing <?php open tag in config example

samsonasik:
Avoid copy paste and got error like in issue:

- https://github.com/rectorphp/rector/issues/8480


TomasVotruba:
Looks good :+1:

Issue #2038, opened 2024-02-09
file permission problem in online demo


staabm:
I tried submitting a new snippet and got a file system error

https://getrector.com/demo/8e474b2d-cc55-4bcd-b87c-5de7ae348fb5

```
 Rector run Failed:
In FileSystem.php line 125: Unable to write file '/var/www/public/rector.php'. Failed to open stream: P ermission denied process [-n|--dry-run] [-a|--autoload-file AUTOLOAD-FILE] [--no-progress-bar] [--no-diffs] [--output-format OUTPUT-FORMAT] [--debug] [--memory-limit MEMORY-LIMIT] [--clear-cache] [--port PORT] [--identifier IDENTIFIER] [--xdebug] [--] [...] 
```

![grafik](https://github.com/rectorphp/rector/assets/120441/3c3dba20-15a0-49dd-ac14-cb30e7e1e75c)

PR #2045, opened 2024-02-12
Validate at least 1 rector rule on Demo page

samsonasik:
@staabm this should fix https://github.com/rectorphp/getrector-com/issues/2038

it is not perfect, but will make less notice

### DIFF
--- a/tests/Controller/DemoControllerTest.php
+++ b/tests/Controller/DemoControllerTest.php
@@ -71,6 +71,11 @@ final class DemoControllerTest extends AbstractTestCase
             'rector_config' => 'PHP config should not include func call',
         ]];
 
+        // Add no rule in config
+        yield ['<?php echo "test"; ?>', '<?php $rectorConfig->removeUnusedImports(); ?>', [
+            'rector_config' => 'PHP config should include at least 1 rector rule',
+        ]];
+
         // valid PHP syntaxes
         yield ['<?php', '<?php', []];
     }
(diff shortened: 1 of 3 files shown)


samsonasik:

stefantalen:
ping @TomasVotruba

Was trying to validate a rule on the demo environment but encountered the error mentioned in https://github.com/rectorphp/getrector-com/issues/2038

samsonasik:
Rebased 👍

tests/Controller/DemoControllerTest.php
```diff
@@ -71,6 +71,11 @@ public static function provideTestFormSubmitData(): Iterator
             'rector_config' => 'PHP config should not include func call',
         ]];
 
+        // Add no rule in config
+        yield ['<?php echo "test"; ?>', '<?php $rectorConfig->removeUnusedImports(); ?>', [
+            'rector_config' => 'PHP config should include at least 1 rector rule',
```
staabm, line 76: maybe instead of "bugging" the user with a error, would it be possible to add a noop rule automatically?

TomasVotruba:
Thank you :+1:

tests/Controller/DemoControllerTest.php
samsonasik: this is the simplest solution I can found right now, add new rule in the config seems requires more complex refactor to detect type and position.

staabm: thank you!

PR #2050, opened 2024-02-13
chore(doc): remove not found demo path

94noni:
hey, imho its better to keep it clean of a 404 notif when devs arrived on the page


samsonasik (approved):
Thank you, this will reduce maintenance update when database cleared up.

PR #2093, opened 2024-02-27
[Demo Validation] Add ShellExecRule rule to avoid `` execution operator usage on rector config input

samsonasik:
Avoid user input danger execution operator as `ShellExec`

```php
<?php `dangerous command`; ?>
```

Ref https://www.php.net/manual/en/language.operators.execution.php


samsonasik:

PR #2117, opened 2024-03-08
fix: typo with prepared sets name

shakaran:


TomasVotruba:
Awesome, thank you for the fix :+1:

Issue #2223, opened 2024-05-03
Error on /ast/100000 when {id} not exists


samsonasik:
when id not exists, it got error: 

```
Rector\Website\PhpParser\NodeResolver\FocusedNodeResolver::focus(): Return value must be of type PhpParser\Node, null returned
```

![Screenshot 2024-05-03 at 12 41 09](https://github.com/rectorphp/getrector-com/assets/459648/58e350ea-0232-4ea6-8e63-d52ba1bd1806)

PR #2224, opened 2024-05-03
Fallback to dump all nodes instead of error on /ast/100000 when id is not exists

samsonasik:
**Before**

Screenshot 2024-05-03 at 12 41 09

**After**

Screenshot 2024-05-03 at 12 47 03

Closes https://github.com/rectorphp/getrector-com/issues/2223


samsonasik:

PR #2230, opened 2024-05-04
Resolve target node to its Stmt holder for AttributeGroup and Attribute on AstDetailController

samsonasik:


samsonasik:

PR #2277, opened 2024-05-26
[Docs] Add note for withPreparedSets() on php 7.x

samsonasik:
On php 7.x, the named argument is not yet exists, this update doc to clarify it.

Ref https://github.com/rectorphp/rector/issues/8653
/cc @orionseye

### DIFF
--- a/resources/docs/introduction.md
+++ b/resources/docs/introduction.md
@@ -43,6 +43,14 @@ return RectorConfig::configure()
     ->withPreparedSets(deadCode: true);
 ```
 
+> If you're on PHP 7.x, you need to passed value argument instead, on `deadCode` argument set as first argument:
+> ```php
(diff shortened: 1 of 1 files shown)


samsonasik:
Using diff for better see the different:

Screenshot 2024-05-26 at 14 44 18

samsonasik:
Ready to merge 👍

TomasVotruba:
I think calling the methods like this:

```php
->withPreparedSets(true, false, true, true);
```
could be a bit unreadable :) I'd recommend to use `withSets([...])` instead, untill the project is upgraded to PHP 8.

samsonasik:
Ok, updated with `withSets()`

TomasVotruba:
Thank you :clap:

PR #2299, opened 2024-06-04
[filer] add diffs in livewire - step 3

TomasVotruba:
Everything works well :) except when the component is re-rendered, the hilightjs is lost.

I've triedy:

- using `@script`, doesn't work
- using `addEventListener` with various `livewire:*` events, doesn't work

[Screencast from 2024-06-04 22-12-43.webm](https://github.com/rectorphp/getrector-com/assets/924196/74aaff55-ea27-473f-900f-b42b53c0dfac)

Any ides how to make hljs re-atach after component is changed?

Olá @ijpatricio, I know we've done something like this together, but I don't have the project anymore. Any tips? :)

### DIFF
--- a/resources/views/base.blade.php
+++ b/resources/views/base.blade.php
@@ -19,14 +19,15 @@
 
         @include('_snippets/layout/footer')
 
+        <!-- Google tag (gtag.js) -->
+        <script async src="https://www.googletagmanager.com/gtag/js?id=UA-46082345-2"></script>
         <script>
-            ga=function(){ ga.q.push(arguments) };
-            ga.q=[];
-            ga.l=+new Date;
-            ga('create', 'UA-46082345-2', 'auto');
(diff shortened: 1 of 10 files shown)


resources/views/base.blade.php
```diff
@@ -19,17 +20,29 @@
...
+
+        <script>
+            // Listen for events dispatched from Livewire components...
+            Livewire.on('rules-filtered', () => {
+                document.querySelectorAll('pre code.language-diff').forEach((block) => {
+                    hljs.highlightBlock(block);
+                });
+
+                // alert('555');
+            })
```
TomasVotruba, line 45: This part

TomasVotruba: Fixed in https://github.com/rectorphp/getrector-com/pull/2300 :clap:

PR #2300, opened 2024-06-04
Feat: hljs after re render

MrPunyapal:
added setTimeout with 0 to run this function after rendering so that it works properly. (POV: I faced this kind of issue a lot after the new update)

also updated `highlightBlock` to `highlightElement` to remove the deprecation warning.

https://github.com/rectorphp/getrector-com/assets/53343069/b6399010-90d5-44b8-bc92-83d94498d8b3

### DIFF
--- a/resources/views/_snippets/styles.blade.php
+++ b/resources/views/_snippets/styles.blade.php
@@ -23,8 +23,8 @@
 
 <script>
     document.addEventListener('DOMContentLoaded', (event) => {
-        document.querySelectorAll('pre code.language-bash, pre code.language-php, pre code.language-yaml, pre code.language-diff, pre code.language-json').forEach((block) => {
-            hljs.highlightBlock(block);
+        document.querySelectorAll('pre code.language-bash, pre code.language-php, pre code.language-yaml, pre code.language-diff, pre code.language-json').forEach((element) => {
+            hljs.highlightElement(element);
         });
     });
 </script>
--- a/resources/views/base.blade.php
+++ b/resources/views/base.blade.php
@@ -36,12 +36,14 @@
 
         <script>
             // Listen for events dispatched from Livewire components...
-            Livewire.on('rules-filtered', () => {
-                document.querySelectorAll('pre code.language-diff').forEach((block) => {
-                    hljs.highlightBlock(block);
+            document.addEventListener('DOMContentLoaded', function () {
+                document.addEventListener('rules-filtered', () => {
+                    setTimeout(() => {
+                        document.querySelectorAll('pre code.language-diff').forEach((element) => {
+                            hljs.highlightElement(element);
+                        });
+                    }, 0);
                 });
-
-                // alert('555');
             })
         </script>
     </body>


resources/views/base.blade.php
```diff
@@ -36,12 +36,14 @@ function gtag(){dataLayer.push(arguments);}
 
         <script>
             // Listen for events dispatched from Livewire components...
-            Livewire.on('rules-filtered', () => {
-                document.querySelectorAll('pre code.language-diff').forEach((block) => {
-                    hljs.highlightBlock(block);
+            document.addEventListener('DOMContentLoaded', function () {
+                document.addEventListener('rules-filtered', () => {
+                    setTimeout(() => {
```
SanderMuller, line 41: I think https://developer.mozilla.org/en-US/docs/Web/API/Window/requestAnimationFrame is the recommended function to do something after rerender (see for some very interesting insights into the JS event loop https://www.youtube.com/watch?v=cCOL7MC4Pl0)

TomasVotruba:
Blazing fast :open_mouth: :clap:  Thank you, I'm looking into it..

resources/views/_snippets/styles.blade.php
```diff
@@ -23,8 +23,8 @@
 
 <script>
     document.addEventListener('DOMContentLoaded', (event) => {
-        document.querySelectorAll('pre code.language-bash, pre code.language-php, pre code.language-yaml, pre code.language-diff, pre code.language-json').forEach((block) => {
-            hljs.highlightBlock(block);
+        document.querySelectorAll('pre code.language-bash, pre code.language-php, pre code.language-yaml, pre code.language-diff, pre code.language-json').forEach((element) => {
+            hljs.highlightElement(element);
```
TomasVotruba, line 27: Thanks for addressing this :clap: . 
I've seen some error in console log, but didn't understand what to do about it

resources/views/base.blade.php
```diff
@@ -36,12 +36,14 @@ function gtag(){dataLayer.push(arguments);}
 
         <script>
             // Listen for events dispatched from Livewire components...
-            Livewire.on('rules-filtered', () => {
-                document.querySelectorAll('pre code.language-diff').forEach((block) => {
-                    hljs.highlightBlock(block);
+            document.addEventListener('DOMContentLoaded', function () {
+                document.addEventListener('rules-filtered', () => {
```
TomasVotruba, line 40: I see, so the trick was about adding one event listener for dom, + nested one for a custom event

resources/views/base.blade.php
TomasVotruba: @SanderMuller Thanks for sharing :+1: 
Indeed the `requestAnimationFrame()` does the refresh without blinking from pure text to hljs colored one :+1:

resources/views/base.blade.php
MrPunyapal: Also setTimeout(callback, 0);

Which ensures the morphing is done and we have new html in the dom.

resources/views/base.blade.php
TomasVotruba: @SanderMuller I've checked it and will use it as it solves the blinking, thanks for sharing :+1:

MrPunyapal: @SanderMuller thanks for sharing it. I leaned something new and it's time to change some codebase of mine. ✅

PR #2301, opened 2024-06-04
Run hljs after requestAnimationFrame

SanderMuller:


resources/views/base.blade.php
```diff
@@ -35,14 +35,16 @@ function gtag(){dataLayer.push(arguments);}
         @livewireScripts
 
         <script>
-            // Listen for events dispatched from Livewire components...
-            Livewire.on('rules-filtered', () => {
-                document.querySelectorAll('pre code.language-diff').forEach((block) => {
-                    hljs.highlightBlock(block);
+            document.addEventListener('livewire:init', () => {
```
SanderMuller, line 38: See https://livewire.laravel.com/docs/events#using-javascript-to-interact-with-events

TomasVotruba: I guess this is different syntax for same calls. The double event nesting is what I missed :) 

But the `requestAnimationFrame()` comes handy, thakns for sharing :+1:

TomasVotruba:
Merged in https://github.com/rectorphp/getrector-com/commit/fd4d23588f12c1eaf6ca3fdf4ea3e2054ef5f7fa

resources/views/base.blade.php
SanderMuller: You might want to still consider using the `'livewire:init'` event over `DOMContentLoaded`, since `'livewire:init'` is fired after Livewire is started, `DOMContentLoaded` can fire before Livewire is started

TomasVotruba: I'll stay with current version to put less coupled knowledge. This way would more likely work in Livewire 4 as well

PR #2334, opened 2024-06-17
Turn ast page to Livewire component

TomasVotruba:


samsonasik:
@TomasVotruba click target node seems refresh to default detail page instead of target node https://getrector.com/ast/d0870f87-523f-44c5-8fbb-a95ccd9e9820/2

TomasVotruba:
I'll look into it...

PR #2359, opened 2024-06-23
[Play and Learn] Add "custom-rule" page for online sharing of custom rules

TomasVotruba:
Just added new form to try out Rector rules on-line. Great for sharing and learning :) 

[Screencast from 2024-06-24 23-17-27.webm](https://github.com/rectorphp/getrector-com/assets/924196/c376ff89-06ec-48a9-a49d-5a9d07fa3963)

Available only in dev environment for now.


resources/default-form-data/custom-rule/CustomRuleRector.php
```diff
@@ -0,0 +1,30 @@
...
+        return [String_::class];
+    }
+
+    /**
+     * @param String_ $node
+     */
+    public function refactor(Node $node)
+    {
+        // change to number
+        return new LNumber(100);
```
samsonasik, line 24: is the page allow user to try their own rule? beware of user add dangerous func call, eg: `shell_exec`, `system`, etc.

this was the incident because of load rule config at demo page that user uses `shell_exec` func call, that's why the rule is added to not allow add func call on config box:

- https://github.com/rectorphp/getrector-com/pull/1967


Screenshot 2024-06-24 at 19 37 40

TomasVotruba: Yes, so people share and try rules easily.

The rule content is evaluated the same way Rector demo config is. So all the forbidden methods are checked here too. Do you think that's good enough?

samsonasik: Rector box config rule currently limit all `FuncCall_` for safest belt, which if we apply on this page, may make create custom rule hard to create as user may rely on native PHP function.

We probably needs list of "safe" functions :

- utilize `PureFunctionDetector::detect()` from Rector for complete list.
- array functions (side effect only on current array, eg: array_pop, etc)
- etc

That needs to be test carefully locally test first before publish online, probably mark this page "dev" only first to test locally.

routes/web.php
```diff
@@ -54,6 +57,11 @@
 Route::get('demo', DemoController::class);
 Route::post('process-demo', ProcessDemoFormController::class);
 
+Route::get('custom-rule/{uuid}', CustomRuleDetailController::class)
+    ->whereUuid('uuid');
+Route::get('custom-rule', CustomRuleController::class);
+Route::post('process-custom-rule', ProcessCustomRuleFormController::class);
```
samsonasik, line 63: this need to be tested locally first for "dev" only try, I checked locally, using this seems work:

```php
if ((int) env('APP_DEBUG') === 1) {
    // ...
}
```

per `.env.prod` for server and `.env.dist` for local

https://github.com/rectorphp/getrector-com/blob/6550ad66a3ff621b211208767ef85f9ab3cdfdd7/.env.prod#L2

https://github.com/rectorphp/getrector-com/blob/6550ad66a3ff621b211208767ef85f9ab3cdfdd7/.env.dist#L2

not sure why `APP_ENV` show as "local"

samsonasik:
Rebase is needed as I add new IncludeRule at:

- https://github.com/rectorphp/getrector-com/pull/2363

TomasVotruba:
First version is ready, let's ship it :) only in dev for now

PR #2404, opened 2024-07-06
[find-rule] Group more nodes under one select, property + property fetch etc

TomasVotruba:


resources/views/livewire/rector-filter-component.blade.php
```diff
@@ -71,7 +71,9 @@ class="form-control d-inline"
             >
                 <div class="card-body pt-0 pb-3 ps-4 pe-4">
                     <div style="float: right" class="mt-4">
-                        <a href="{{ action(\App\Controller\RuleDetailController::class, ['slug' => $filteredRule->getSlug()]) }}">Rule detail</a>
+                        @env('dev')
+                            <a href="{{ action(\App\Controller\RuleDetailController::class, ['slug' => $filteredRule->getSlug()]) }}">Rule detail</a>
+                        @endenv
```
TomasVotruba, line 76: Ref https://github.com/rectorphp/getrector-com/issues/2405

I'll reword the coding standard run from process, that is most likely causing the error,  to php-parser printer itself

PR #2414, opened 2024-07-08
[Validation] Part 1: allow non-side effect func calls on rector config web page

samsonasik:


src/Validation/Rules/ForbiddenFuncCallRule.php
```diff
@@ -19,17 +25,51 @@ public function validate(string $attribute, mixed $value, Closure $fail): void
         $parserFactory = new ParserFactory();
         $parser = $parserFactory->create(ParserFactory::PREFER_PHP7);
 
+        $lazyContainerFactory = new LazyContainerFactory();
+        $rectorConfig = $lazyContainerFactory->create();
+        $signatureMapProvider = $rectorConfig->make(PHPStanServicesFactory::class)->getByType(
```
TomasVotruba, line 30: I think using PHPStan container factory here is overkill. It would spawn so many services and full PHPStan container outside Rector.

Instead, we should use simple function name check 👍

samsonasik: `PHPStanServicesFactory` is always executed when running rector demo, as we require phpstan service internally, so should be ok to be called here as well.

There are many side effect functions, we, of course, can use more complete `PureFunctionDetector`, but that will require MutatingScope to be applied into Node, which doesn't make sense on the validation.

This rule won't be executed on normal open web page, only executed when POST demo executed, so should be ok as is.

On next step, we can identify more "safe" side effect that allowed, eg: `array_push()`, `array_pop`, etc.

TomasVotruba: Keep it simple: https://github.com/staabm/phpstan-src/blob/b051b9930bfd1a57d7108703c8919acebefbd717/src/Reflection/SignatureMap/functionMetadata.php

samsonasik: there seems a magic things to load obvious dangerous native php function in the function metadatas, I can't find the `exec` definition for `'hasSideEffects' => true`, it only verifiable via `SignatureMapProvider` that compile the data, possibly compiled with what phpstorm/stubs provided.

samsonasik: First investigation, it seems `exec` seems not registered as `hasSideEffects => true`, on my code handling, it fallback on `else` part which make it true:

```php
                    if ($signatureMapProvider->hasFunctionMetadata($name)) {
                        $hasSideEffects = TrinaryLogic::createFromBoolean(
                            $signatureMapProvider->getFunctionMetadata($name)['hasSideEffects']
                        );
                    } else {
                        // possibly unknown
                        $hasSideEffects = TrinaryLogic::createYes();
                    }
```

On phpstan code base, it fallback to `TrinaryLogic::createMaybe()`

samsonasik: Also updated to sync with what phpstan does https://github.com/rectorphp/getrector-com/pull/2414/commits/3cdb26d0686886a93404fa84db028da8f0dd018f 👍

TomasVotruba (approved):

TomasVotruba:
Looks simple and effective 🙏👌

PR #2420, opened 2024-07-11
[Validation] Make HasRectorRule use more reliable Rector container factory

TomasVotruba:


src/Request/RectorRunFormRequest.php
```diff
@@ -25,38 +25,38 @@ public function authorize(): bool
     public function rules(): array
     {
         $shortPhpContentsRule = $this->make(ShortPhpContentsRule::class);
-        $validPhpSyntaxRule = $this->make(ValidPhpSyntaxRule::class);
+        $validAndSafePhpSyntaxRule = $this->make(ValidAndSafePhpSyntaxRule::class);
+
+        // @todo list forbidden functions? merge into @see ValidAndSafePhpSyntaxRule
```
samsonasik, line 30: I am doing this on wip PR:

- https://github.com/rectorphp/getrector-com/pull/2414

TomasVotruba: Awesome :+1: I'll leave that to you then

phpunit.xml
```diff
@@ -14,5 +14,6 @@
     </testsuites>
     <php>
         <env name="APP_ENV" value="test"/>
+        <env name="APP_BASE_PATH" value="."/>
```
TomasVotruba, line 17: This is important path to differ from Rector container collision :+1:

src/Validation/Rules/HasRectorRule.php
```diff
@@ -7,64 +7,58 @@
...
-        $parserFactory = new ParserFactory();
-        $parser = $parserFactory->create(ParserFactory::PREFER_PHP7);
-
         try {
-            $stmts = $parser->parse($value);
-            $nodeFinder = new NodeFinder();
+            $filesystem = new Filesystem();
+
+            $configFilePath = sys_get_temp_dir() . '/temp-rector-config.php';
+            $filesystem->dumpFile($configFilePath, $value);
```
TomasVotruba, line 32: This might be problematic spot on production

TomasVotruba: Just deployed and looks good:

Screenshot from 2024-07-11 23-57-34

samsonasik: this seems cause possible error https://github.com/rectorphp/getrector-com/issues/2438, also for consecutive request, the file match replace each other, so it may need random name, and remove after it usage.

samsonasik: https://github.com/rectorphp/getrector-com/pull/2446

PR #2435, opened 2024-07-14
[DX] Move diffs right to the code

TomasVotruba:
## Before

The longer the content, the more scrolling to find the diff :face_exhaling: 

Screenshot from 2024-08-02 10-25-22

---

## After

Diff is right in the code, so no more scrolling.

Screenshot from 2024-07-15 01-31-03

### DIFF
--- a/src/Entity/AbstractRectorRun.php
+++ b/src/Entity/AbstractRectorRun.php
@@ -4,11 +4,13 @@ declare(strict_types=1);
 
 namespace App\Entity;
 
+use App\Demo\ValueObject\DiffSnippet;
 use App\Enum\Comment;
 use App\Utils\FileDiffCleaner;
 use JsonSerializable;
 use Nette\Utils\Strings;
 use Symfony\Component\Uid\Uuid;
+use Webmozart\Assert\Assert;
 
 abstract class AbstractRectorRun implements JsonSerializable
 {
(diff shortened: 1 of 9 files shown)


TomasVotruba:
Note: add light + dark theme of diffs .)

src/Entity/AbstractRectorRun.php
```diff
@@ -27,9 +29,33 @@ public function getUuid(): Uuid
...
+    /**
+     * @return DiffSnippet[]
+     */
+    public function getDiffSnippets(): array
+    {
+        $diffSnippets = [];
+
+        $contentDiffsRaw = Strings::split($this->getContentDiff(), '#@@ #');
+        foreach (array_filter($contentDiffsRaw) as $contentDiffRaw) {
+            [$lines, $content] = Strings::split($contentDiffRaw, "# @@\n#");
```
samsonasik, line 41: it seems cause error:

```
 Undefined array key 1 
```

Screenshot 2024-08-02 at 18 45 25

with code example:

```php
<?php

namespace Miofair\LayoutBundle\Tests\Form\Type;

use PHPUnit\Framework\MockObject\MockObject;
use PHPUnit\Framework\TestCase;

class OnCallbackTest extends TestCase
{
    private (\Exception&MockObject)|null $mock = null;

    protected function setUp(): void
    {
        $this->mock = $this->createMock(\Exception::class);
    }

    public function testConfigureOptions(): void
    {
        $matcher = static::exactly(6);
        $this->mock
            ->expects($matcher)
            ->method('__toString')
            ->willReturnCallback(function (){
                return $this->mock;
            })
        ;
    }
}
```

and code config:

```php
<?php

use Rector\Config\RectorConfig;
use Rector\TypeDeclaration\Rector\Property\TypedPropertyFromAssignsRector;

return RectorConfig::configure()
    // A. whole set
 //   ->withPreparedSets(typeDeclarations: true)
    // B. or few rules
    ->withRules([
        TypedPropertyFromAssignsRector::class
    ]);
```

samsonasik: It happen when content result is:

```
"// no change"
```

samsonasik: https://github.com/rectorphp/getrector-com/pull/2488

PR #2446, opened 2024-07-17
[Validation] Randomize and delete after unused for generated rector config on HasRectorRule

samsonasik:
this ensure no overlapped consecutive request on rector demo page, also, if generated config no longer used, directly removed.


samsonasik:

TomasVotruba:
Looks good :+1:

PR #2447, opened 2024-07-17
[Demo] Remove generated directory config on DemoRunner::processFilesContents()

samsonasik:
DemoRunner currently always create new directory without removal, and only remove files inside it, that cause a lot of empty directories:

Screenshot 2024-07-17 at 22 19 43

this patch fix it.


samsonasik:

TomasVotruba:
Thanks :+1:

PR #2449, opened 2024-07-17
Make RectorConfig as a service for reuse it on validation services

samsonasik:


samsonasik:

src/Providers/AppServiceProvider.php
```diff
@@ -16,6 +18,11 @@ final class AppServiceProvider extends ServiceProvider
 {
     public function register(): void
     {
+        $lazyContainerFactory = new LazyContainerFactory();
+        $rectorConfig = $lazyContainerFactory->create();
+
+        $this->app->singleton(RectorConfig::class, fn (): RectorConfig => $rectorConfig);
+
```
TomasVotruba, line 25: RectorConfig is a value object, so it should never be a service. It's like framework configuration, created and used just once. This could lead to very weird bugs while using different configs for validation and for Rector run :)

samsonasik: I want to use the service in `ForbiddenFuncCallRule` on PR:

- https://github.com/rectorphp/getrector-com/pull/2414

it seems can't be redundancy, as re-create `$lazyContainerFactory->create()` in multiple classes cause error `command not found`

PR #2453, opened 2024-07-19
[validation] Add ForbiddenCallLikeRule

samsonasik:
On demo page, verify config file consist of allowed CallLike:

```php
new \PHPStan\Type\MixedType();
```

but not dangerous CallLIke:

```php
(new Nette\Utils\FileSystem)->write('test.php', 'test');
```


samsonasik:

samsonasik:
I will add more test for demo page fluent:

```php
<?php

use Rector\Config\RectorConfig;
use Rector\TypeDeclaration\Rector\Property\TypedPropertyFromAssignsRector;

return RectorConfig::configure()
    // A. whole set
    ->withPreparedSets(typeDeclarations: true)
    // B. or few rules
    ->withRules([
        TypedPropertyFromAssignsRector::class
    ]);

```

samsonasik:

tests/Controller/DemoControllerTest.php
```diff
@@ -61,21 +62,86 @@ public function testValidRequest(): void
         $this->assertFalse($testResponse->isServerError());
     }
 
-    public function testIncludeNonDangerousFuncCallRequest(): void
+    #[RunInSeparateProcess]
```
TomasVotruba, line 65: This should not be used, as hiding a complex behavior of container. Could you remove it?
What is the error then?

samsonasik: checking assert on test response seems by session, once defined, it reused, so I used it, if it removed, it got invalid result:

```
There was 1 failure:

1) App\Tests\Controller\DemoControllerTest::testValidStrStartsWith
Session has unexpected errors: 

{
    "default": [
        "PHP config should not include side effect func call"
    ]
}
Failed asserting that true is false.

/Users/samsonasik/www/getrector.org/vendor/laravel/framework/src/Illuminate/Testing/TestResponseAssert.php:45
/Users/samsonasik/www/getrector.org/vendor/laravel/framework/src/Illuminate/Testing/TestResponse.php:1409
/Users/samsonasik/www/getrector.org/tests/Controller/DemoControllerTest.php:75
```

TomasVotruba: Tests should reset session then, to start with clean plate.

samsonasik: Ok, on PHPUnit, it seems session issue fixed by change `include_once` to `include` on get function metadata https://github.com/rectorphp/getrector-com/pull/2453/commits/403efd5ecfec4596960e293e624cd90f43653dd1

samsonasik:
All checks have passed 🎉 @TomasVotruba I think it is ready. It is not perfect,but good enough for starter ;)

TomasVotruba (approved):

TomasVotruba:
Looks good, thanks :+1:

PR #2454, opened 2024-07-19
[validation] Handle provide withPaths() on demo page which pointed to invalid paths

samsonasik:
**Before**

Screenshot 2024-07-19 at 22 26 10

**After**

Screenshot 2024-07-19 at 22 25 56

### DIFF
--- a/src/Validation/Rules/HasRectorRule.php
+++ b/src/Validation/Rules/HasRectorRule.php
@@ -75,6 +75,10 @@ final class HasRectorRule implements ValidationRule
                     throw new ShouldNotHappenException('PHP config should have valid method name, you may have typo');
                 }
 
+                if (str_ends_with($message, 'does not exist.')) {
+                    throw new ShouldNotHappenException('PHP config should have valid paths');
+                }
+
                 throw new ShouldNotHappenException('Expected config should return callable RectorConfig instance');
             }
         }
--- a/src/Validation/Rules/HasRectorRule.php
+++ b/src/Validation/Rules/HasRectorRule.php
@@ -69,17 +69,8 @@ final class HasRectorRule implements ValidationRule
         foreach ($configFiles as $configFile) {
             try {
                 $rectorConfig->import($configFile);
-            } catch (Throwable $e) {
-                $message = $e->getMessage();
-                if (str_starts_with($message, 'Call to undefined method')) {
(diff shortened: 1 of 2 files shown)


TomasVotruba:
I wonder, if there should be allowed any paths at all, as we only run single file.

samsonasik:
is error message needs to be changed? whats better wording for it?

TomasVotruba:
I'd forbid running `withPaths()` completelly, as no paths except provided file are used.

samsonasik:
The issue seems due to too early usage of `$rectorConfig->import()` on `HasRectorRule`

https://github.com/rectorphp/getrector-com/blob/91a12c0444d59acdd271a1705309b2468bc93a16/src/Validation/Rules/HasRectorRule.php#L71

so override by path command is not yet executed.

samsonasik:
The Assert::filexists on rector-src seems need to be moved to resolving

https://github.com/rectorphp/rector-src/blob/0beeb6fa8beb9fd3833c7de4ba3b3a0cdd408c3b/src/Configuration/ConfigurationFactory.php#L120

samsonasik:
It seems verbose:

```php
throw new ShouldNotHappenException($t->getMessage());
```

On fatal error is better on this case, as it may various kind of different message, user will already clear see what needs to do.

samsonasik:
I updated to use its fatal error mesaage into display message https://github.com/rectorphp/getrector-com/pull/2454/commits/4eea5b0c2c8b2d9feb872f7558f5bc0fb29bf1d6

samsonasik:
use verbose message pulled from the throwable error seems naturally better 👍 

Screenshot 2024-07-20 at 12 08 29

TomasVotruba:
That could lead to people fixing the path to another. But there is only single file path that is being processed. We create it in the background. The `withPaths()` should never be used in our demo.

E.g. that's why there are no paths in PHPStan demo: https://phpstan.org/try

samsonasik:
I will check if that possible, for that, we may need a "filter" instead of "validation", eg: `FilterPaths` that clean `withPaths()` from `RectorConfigBuilder` or `paths()` from `RectorConfig`, as in current implementation of `HasRectorRule`, the `$ $rectorConfig->import($configFile); ` is called too early

samsonasik:
closing for now.

PR #2457, opened 2024-07-20
Temporary disable ForbiddenCallLikeRule as cause error on getrector.com/demo

samsonasik:
it cause error on server

Screenshot 2024-07-21 at 01 20 44

will check directly on server.


samsonasik:
The fix https://github.com/rectorphp/getrector-com/pull/2458

PR #2490, opened 2024-08-03
[validation] Add ShortPhpConfigRule

samsonasik:
Currently, there short maximum lines = 100 is on php contents, but on php config, it doesn't have this validation.

This PR add `ShortPhpConfigRule` for config rule, which the config validation config box, which shared with "Create custom rule" page to avoid too many lines added on the page.

This `ShortPhpConfigRule` have 200 lines maximum.

### DIFF
--- a/src/Validation/Rules/ShortPhpLinesTrait.php
+++ b/src/Validation/Rules/ShortPhpLinesTrait.php
@@ -1,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Validation\Rules;
+
+use Closure;
+use Illuminate\Translation\PotentiallyTranslatedString;
+
+/**
+ * @internal
+ * @method int getInputLineLimit()
+ */
+trait ShortPhpLinesTrait
+{
+    /**
+     * @param Closure(string):PotentiallyTranslatedString $fail
+     */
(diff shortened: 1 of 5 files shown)


samsonasik:

src/Validation/Rules/ShortPhpLinesTrait.php
```diff
@@ -0,0 +1,35 @@
...
+namespace App\Validation\Rules;
+
+use Closure;
+use Illuminate\Translation\PotentiallyTranslatedString;
+
+/**
+ * @internal
+ * @method int getInputLineLimit()
+ */
+trait ShortPhpLinesTrait
```
TomasVotruba, line 14: Instead of trait with this `@method` magic mess, use normal service.

samsonasik: sure, updated to use normal service https://github.com/rectorphp/getrector-com/pull/2490/commits/618cbd13acd50c252ecf8cf2d969621d1d9b4675

samsonasik:

TomasVotruba (approved):

TomasVotruba:
Thanks 🙏👍

PR #2513, opened 2024-08-14
Improve installation docs with database setup

tigitz:


samsonasik:
There is first-run.sh, that should do the same

https://github.com/rectorphp/getrector-com/blob/main/first-run.sh

tigitz:
I'm fine with documenting running `first-run.sh` too instead of the actual instructions in the README. 

However in the `first-run.sh`:
```sh
chgrp -R www-data database
chmod -R g+rw database
mkdir -p storage/framework/{cache,sessions,views}

# create the manifest.json file
yarn build
```

these instructions didnt seem to be needed on my side, I guess it's because you're running it through nginx while I use the simpler `artisan serve` documented in the README.md.

If messing with users and folder permissions could be avoided I suppose it's a win.

Let me know :+1:

TomasVotruba:
Looks good to me :+1:

PR #2519, opened 2024-08-16
Add "step-by-step" minitool to visualise rector.php contents in standard upgrade

TomasVotruba:
chrome-capture-2024-8-18


TomasVotruba:
Let's kick of first version :)

PR #2522, opened 2024-08-18
Add links to rector sets from the filter results and rule detail pages

carlos-granados:
When you are searching for rules in the rule finder page and you find a rule which belongs to a set, you may want to know which other rules belong to that set.
In this PR we add links to the red badges with the set names which will take you straight to a list of all the rules that belong to that set

### DIFF
--- a/resources/views/homepage/rule-detail.blade.php
+++ b/resources/views/homepage/rule-detail.blade.php
@@ -80,9 +80,7 @@
                             SETS:&nbsp;
 
                             @foreach ($ruleMetadata->getSets() as $set)
-                                <a href="{{ action(\App\Controller\FilterRectorController::class, ['rectorSet' => $set->getSlug()]) }}">
-                                    <span class="badge bg-danger">{{ $set->getName() }}</span>
-                                </a>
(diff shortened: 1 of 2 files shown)


TomasVotruba:
Thanks 🙏 

Could you share before/after screen? It's easier to review for me without testing branch locally

carlos-granados:
It is difficult to show this new functionality using static pictures, so I recorded a small video instead, hope that you can see it.

While recording this I realised that the spaces between the badges were being rendered as links as well, so I modified the code to avoid this. I had to add some really long lines but there does not seem to be a way to remove whitespace in blade templates.


https://github.com/user-attachments/assets/27f0115f-dfe1-43da-80b3-325c56df9199

TomasVotruba:
Wow, even better :pray:  let's :ship:  it!

<br>

I love these videos :) it's also perfect way to show people new features on Twitter.

Btw, how did you record it? I look for a Linux solution that could do Twitter-compatible recording like this.

carlos-granados:
@TomasVotruba I used a Mac to record it, with Macs it's really easy, QuickTime has a screen recording option

PR #2556, opened 2024-08-30
[find-rule] Add Laravel sets and improve group select

TomasVotruba:

PR #2579, opened 2024-09-09
Laravel set list

peterfox:
Added to the Laravel set list for https://github.com/rectorphp/getrector-com/pull/2556

### DIFF
--- a/src/RuleFilter/Bridge/LaravelSetProvider.php
+++ b/src/RuleFilter/Bridge/LaravelSetProvider.php
@@ -8,13 +8,40 @@ use App\Enum\FindRule\GroupName;
 use Rector\Set\Contract\SetProviderInterface;
 use Rector\Set\ValueObject\Set;
 use RectorLaravel\Set\LaravelSetList;
+use RectorLaravel\Set\Packages\Livewire\LivewireSetList;
 
 /**
- * @todo should be part of the Laravel package, move later
  * @see https://github.com/driftingly/rector-laravel/tree/main/config/sets
  */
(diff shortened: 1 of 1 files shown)


src/RuleFilter/Bridge/LaravelSetProvider.php
```diff
@@ -8,13 +8,40 @@
 use Rector\Set\Contract\SetProviderInterface;
 use Rector\Set\ValueObject\Set;
 use RectorLaravel\Set\LaravelSetList;
+use RectorLaravel\Set\Packages\Livewire\LivewireSetList;
 
 /**
- * @todo should be part of the Laravel package, move later
  * @see https://github.com/driftingly/rector-laravel/tree/main/config/sets
  */
 final class LaravelSetProvider implements SetProviderInterface
```
TomasVotruba, line 16: Awesome!

While we're at it, this `LaravelSetProvider` class should be part of https://github.com/driftingly/rector-laravel/ code.

Could you add it there, so we can just re-use the dependency?

peterfox: Yes, of course. My only question is about `GroupName::LARAVEL`, will it be fine if the value of `laravel` is just used as the class is in the getrector repository?

TomasVotruba:
Looking into this today :eyes:

TomasVotruba:
Rebased and merged in https://github.com/rectorphp/getrector-com/pull/2587

Thank you :+1: 

Deploy coming soon...

PR #2581, opened 2024-09-09
[post] Joining OSS Pledge

TomasVotruba:
Ref https://osspledge.com/join/


.meta/oss-pledge.json
```diff
@@ -0,0 +1,22 @@
+{
```
samsonasik, line 1: is this file needs to under public/ directory so it fetchable?

TomasVotruba: It's available via Github repo link 👍

PR #2584, opened 2024-09-10
Improving Rector Performance blog post

carlos-granados:
@TomasVotruba as agreed here is my blog post about Improving Rector Performance. Please take a look and let me know if there is anything that needs changing. Please also let me know once this is live. Thanks!!

### DIFF
--- a/resources/blog/posts/2024/2024-09-11-improving-rector-performance.md
+++ b/resources/blog/posts/2024/2024-09-11-improving-rector-performance.md
@@ -9,7 +9,7 @@ perex: |
 
 <br>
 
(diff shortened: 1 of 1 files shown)


carlos-granados:
It seems that it is just that the file preview cannot show this properly, hopefully it will show correctly once published

carlos-granados:
I just tested this by launching a local version of the site and everything seems to be displaying ok, so I hope everything is fine 😄

TomasVotruba:
Looking into this :) don't worry about the file preview, I'll check it

TomasVotruba:
I've just read it and the  post looks awesome. It explains the issue, the solution and benefits.
Nice and clear :clap:  :partying_face:

PR #2585, opened 2024-09-10
add author picture and name for socials

TomasVotruba:


TomasVotruba:
Follow up to https://github.com/rectorphp/getrector-com/pull/2584

PR #2587, opened 2024-09-18
Add Laravel sets to "find rule" page

TomasVotruba:
Looks good :clap: 

Screenshot from 2024-09-18 11-01-03


samsonasik:
When one of the Laravel rule clicked, it seems back to `/find-rule` instead of detail page, step to reproduce:

1. Open https://getrector.com/find-rule?activeRectorSetGroup=laravel
2. Click one of the rule listed.
3. It should go to detail rule, it currently back to `/find-rule`

PR #2639, opened 2024-10-20
Bug missing negation on blog post 5 Tricks to Write Better Custom Rules

samsonasik:
tested at https://getrector.com/custom-rule/e2629342-0432-4b77-aaa2-8f8c5808b87d


TomasVotruba:
Thanks 👍

PR #2688, opened 2024-11-21
Bump to PHPStan 2 and PHPParser 5

samsonasik:


samsonasik:

PR #2800, opened 2025-01-09
Fix Laravel rule filtering

GeniJaho:
When filtering the rules in the [Find Rule](https://getrector.com/find-rule?activeRectorSetGroup=laravel) page, the Laravel group only shows 38 rules, while there are 76 available at the moment. That happens because we only show and count the Laravel rules that belong to a set, as with the rest of the Rector groups.
That leaves out most of the Laravel rules.

In this PR we:
- Add an exception when filtering by Group, to check if the rule class starts with `RectorLaravel`
- Add `RuleFilterTest` to cover the edge case
- Use `dev-main` for the Laravel Rector rules, since rules can be available before we do a release
- Add the `LARAVEL` group name and replaces the string literal uses

Let me know please if this is OK or I need to change something.
And thank you for making the code open for the Rector website. :smile:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
     },
     "require-dev": {
         "barryvdh/laravel-ide-helper": "^3.2",
-        "driftingly/rector-laravel": "^2.0",
+        "driftingly/rector-laravel": "dev-main as 2.0",
         "nette/robot-loader": "^4.0",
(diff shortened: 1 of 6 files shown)


TomasVotruba:
I've missed this as CI doesn't report success, my appologies.

Just merging now :+1:  thank you

Issue #2826, opened 2025-01-24
/find-rule got 500 error


samsonasik:
It seems /find-rule now got 500 error page

![Image](https://github.com/user-attachments/assets/8618cae2-9d73-4b42-8c8b-d783cded25e9)

PR #2827, opened 2025-01-24
Fix /find-rule 500 error

samsonasik:
`driftingly/rector-laravel` requirement should use tagged version which downgraded to php 7.2 https://github.com/driftingly/rector-laravel/tree/2.0.2

Fixes https://github.com/rectorphp/getrector-com/issues/2826


samsonasik:
Fixed and deployed 👍

TomasVotruba:
Thanks :pray:

PR #2829, opened 2025-01-25
Fix mobile show "parse" button on /ast page

samsonasik:
**Before**

Screenshot 2025-01-25 at 20 04 18

**After**

Screenshot 2025-01-25 at 20 07 55

I am not good at css, but here we go :)

### DIFF
--- a/resources/views/ast/ast.blade.php
+++ b/resources/views/ast/ast.blade.php
@@ -35,10 +35,11 @@
                     </button>
                 </div>
 
+                <br /><br />
+
             </form>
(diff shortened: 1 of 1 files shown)


samsonasik:
@TomasVotruba ready to merge and try 👍

samsonasik:
@TomasVotruba let's merge and try 👍

TomasVotruba:
Just checking this, but it's wrong :) the button must remain in above the form to save the space. 
I'll look into it

PR #2831, opened 2025-01-27
Add withCodingStyle() info to the docs

carlos-granados:
See https://github.com/rectorphp/rector-src/pull/6695


TomasVotruba:
Thank you :partying_face:

PR #2956, opened 2025-04-14
[demo] Split code and diffs into 2 separate columns to avoid overflow

TomasVotruba:
Current solution uses `position: absolute` that disallows clearing div. That's why code overflows to next element.

The only path nows seems spliiting into 2 columns: one with origin code and 2nd with diffs.

Screenshot From 2025-04-14 15-50-12

Closes https://github.com/rectorphp/getrector-com/issues/2957

### DIFF
--- a/resources/views/_snippets/form/form_diff.blade.php
+++ b/resources/views/_snippets/form/form_diff.blade.php
@@ -8,7 +8,7 @@
 @if ($rectorRun->isSuccessful())
     <div class="col-6" style="float: right;" class="diff-snippet">
         @foreach ($rectorRun->getDiffSnippets() as $diffSnippet)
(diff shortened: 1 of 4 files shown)


samsonasik:
It seems on default example, it cause splitted right bottom

Screenshot 2025-04-15 at 18 02 39

TomasVotruba:
It seems it's a dead end - either we break diff positions, or the diff overflows to text.

<br>

I came up with a creative solution, to tidy up buttons under text - there should be 1 button a page anyway :hugs: 

It looks pretty cool: there is section on the left with all details and actions, and right is space for diffs.

↓



## Before  :older_adult: 

Screenshot From 2025-04-15 16-05-35


---

## After :party:

Screenshot From 2025-04-15 16-05-03

samsonasik:
It seems "Add a test fixture" link is not shown on default example

Screenshot 2025-04-15 at 21 17 17

samsonasik:
Oh I see, create a fixture only doable on 1 rule applied.

Issue #3001, opened 2025-04-30
Menu not clickable on homepage / with mobile


samsonasik:
On mobile, it not clickable on / page

![Image](https://github.com/user-attachments/assets/255579f4-6dc5-4e46-9f3e-d006c9fde6f8)

On other page, it clickable (eg: /demo or /documentation) it seems some css not loaded in / page

samsonasik:
Add `'codeMirror' => true,` on `HomepageController` seems make menu clickable, it seems when `'codeMirror' => true,`, some css/js is loaded for menu as well.

samsonasik:
Step to reproduce locally:

- Execute `php artisan serve`
- Open http://localhost:8000/
- Minimize windows

![Image](https://github.com/user-attachments/assets/255579f4-6dc5-4e46-9f3e-d006c9fde6f8)

- Click menu in header left

samsonasik:
I see, bootstrap is needed to be loaded to make menu work

PR #3002, opened 2025-04-30
Fix bootstrap menu not loaded on homepage / that cause menu not clickable

samsonasik:
Fix menu not clickable on homepage /

Fixes https://github.com/rectorphp/getrector-com/issues/3001


samsonasik:
Fixed and deployed 👍

TomasVotruba:
:+1:

PR #3004, opened 2025-04-30
Fix generated meeting.jpg boy finger

samsonasik:
The meeting.jpg seems image generated, and the boy finger seems stacked

meeting_original

This PR try to fix it generated via www.genspark.ai

This however have **GenSpark watermark** tho 

meeting


samsonasik:
Oh, the generated image is different size, before 768x1024, after 1024x1024. I can regenerate next day if needed as I've used all my free credits on www.genspark.a

samsonasik:
I see, it seems the original is using grok, I will try using grok

TomasVotruba:
Feel free to use yours, I like it! :) 
Just cut out the watermark, so image is clean

samsonasik:
The original image is using "Grok" watermark tho, and retry on grok seems make it worse, I will retry tomorrow on genspark.ai

samsonasik:
I've just resize the picture with possible ratio and remove the watermark 👍 , let's merge 👍

samsonasik:
meeting

samsonasik:
@TomasVotruba I deployed so you can check 👍 , clear the browser cache if it still load old image

PR #3006, opened 2025-04-30
Only show Dark/Light Theme bottom menu when on @isset ($codeMirror)

samsonasik:
Avoid error undefined `theme` javascript error, as only when codeMirror is set. 

```
[localhost:8000:56:22](http://localhost:8000/)
Uncaught TypeError: theme is null
    toggleTheme http://localhost:8000/:72
    onclick http://localhost:8000/:1
```

### DIFF
--- a/resources/views/_snippets/menu_items.blade.php
+++ b/resources/views/_snippets/menu_items.blade.php
@@ -46,7 +46,10 @@
            class="nav-link">Find rule</a>
     </li>
 
+    @isset ($codeMirror)
     <li class="nav-item">
         <a onclick="toggleTheme()" class="nav-link" id="theme_toggle" style="cursor: pointer">Dark/Light Theme</a>
     </li>
+    @endisset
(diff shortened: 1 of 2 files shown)


samsonasik:
Fixed and deployed 👍

TomasVotruba:
This should be available on any page, so people are not confused by disappearing option.

Independent on code mirror itself.

samsonasik:
Then it seems variable id "theme-link" exists seems needed to avoid error:

```
[localhost:8000:56:22](http://localhost:8000/)
Uncaught TypeError: theme is null
    toggleTheme http://localhost:8000/:72
    onclick http://localhost:8000/:1
```

samsonasik:
@TomasVotruba I created PR :

- https://github.com/rectorphp/getrector-com/pull/3007

to re-add it while avoid error `Uncaught TypeError: theme is null`

PR #3007, opened 2025-04-30
Re-add Dark/Light Theme link

samsonasik:
@TomasVotruba per review:

- https://github.com/rectorphp/getrector-com/pull/3006#issuecomment-2842744935

The id="theme-link" needs to be exists to avoid error:

```
[localhost:8000:56:22](http://localhost:8000/)
Uncaught TypeError: theme is null
    toggleTheme http://localhost:8000/:72
    onclick http://localhost:8000/:1
```

### DIFF
--- a/resources/views/_snippets/menu_items.blade.php
+++ b/resources/views/_snippets/menu_items.blade.php
@@ -46,10 +46,8 @@
            class="nav-link">Find rule</a>
     </li>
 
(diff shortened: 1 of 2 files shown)


samsonasik:
Fixed and deployed 👍

TomasVotruba:
Looks good, thank you :+1:

PR #3009, opened 2025-05-01
tv contact

TomasVotruba:
- **fix contact photo with real one**
- **bump deps**


samsonasik:
the woman hand seems stacked

TomasVotruba:
it's hand of the guy on the right, this is real pic :D

public/assets/images/meeting.png
samsonasik, line 1: the woman hand in the middle seems stacked

meeting

samsonasik: I see, the finger is just too big

samsonasik: let me regenerate with genspark

samsonasik: oh, it seems hard to make it more realistic, I am fine with this then

PR #3044, opened 2025-05-14
feat(docs): enhance DocumentationMenuItem validation and file checks

monayemislam:
Enhance `DocumentationMenuItem` class with validation and new method. Added parameter descriptions for constructor, enforced `href` and label non-emptiness, and introduced `hasDocumentation` method to check file existence.


TomasVotruba:
Thanks :+1: 

I've approved CI to see how it runs. Tests needs fix now.

PR #3139, opened 2025-07-02
[Docs] Add docs for troubleshoot preload conflict on phpunit 12+

samsonasik:
@ostrolucky @TomasVotruba this is new documentation section for troubleshooting preload conflict on phpunit 12+

Ref 

- https://github.com/rectorphp/rector/issues/9242
- https://github.com/ostrolucky/rector-rules/pull/44

### DIFF
--- a/resources/docs/troubleshooting-preload.md
+++ b/resources/docs/troubleshooting-preload.md
@@ -1,0 +1,28 @@
+On some dependency with CLI command, eg: PHPUnit 12+, you may got error:
+
+```bash
+PHP Fatal error:  Cannot declare interface PhpParser\NodeVisitor, because the name is already in use in /home/runner/work/rector-rules/rector-rules/vendor/rector/rector/vendor/nikic/php-parser/lib/PhpParser/NodeVisitor.php on line 6
+```
+
+which its own autoload too early loaded, you may need to handle it with register `rector` preload.php under `"autoload-dev" -> "files"` like below:
+
+```json
+{
+    "autoload-dev": {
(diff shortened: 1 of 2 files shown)


samsonasik:

TomasVotruba:
This look a bit hacky. What is the change in PHPUnit 12 that caused this?

ostrolucky:
Unfortunately it seems this change breaks phpstan https://github.com/ostrolucky/rector-rules/actions/runs/16016236306/job/45187044734?pr=44

samsonasik:
I am not sure to be honest, but these 2 lines for too early load is shown here:

- https://github.com/sebastianbergmann/phpunit/blob/7c2e0781d70a2455041ef7e26c46b37bd558b3df/phpunit#L46
- https://github.com/sebastianbergmann/phpunit/blob/7c2e0781d70a2455041ef7e26c46b37bd558b3df/phpunit#L72

even without setup phpunit bootstrap, it will always load `vendor/autoload.php` too early.

That happen only on phpunit 12+, with php 8.3+, so either it phpunit issue, or php 8.3 changed behaviour.

samsonasik:
@ostrolucky in seems in rector `preload.php`, it needs initialized:

```php
if (defined('__PHPSTAN_RUNNING__')) {
    return;
}
```

I will create PR for that.

samsonasik:
@TomasVotruba let's merge this, so it sync with PR:

- https://github.com/rectorphp/rector-src/pull/7040

PR #3159, opened 2025-07-12
Upgrade to use php 8.3

samsonasik:
Since server is using php 8.3 now, I think we can upgrade to php 8.3 for web requirement as well.


samsonasik:
All checks have passed 🎉 @TomasVotruba I think it is ready 👍

TomasVotruba:
Looks great, thanks :clap: :partying_face:

PR #3161, opened 2025-07-13
Improve local dev support on cloud IDEs

codebymikey:
Just some minor devops improvements to aid when running the codebase on a cloud IDE like Github Codespaces or Gitpod.

* Add sqlite3 as a require-dev composer dependency so it's a bit more obvious that the extension is necessary in order to run the site.
* Use relative paths for links and assets, makes it easier to run under a reverse proxy even if a reverse proxy isn't available.
* Address missing favicon.ico link, the favicon currently 404s.
* Enable local reverse proxy support when running on a `dev` environment.


composer.json
```diff
@@ -19,6 +19,8 @@
         "symplify/vendor-patches": "^11.5"
     },
     "require-dev": {
+        "ext-sqlite3": "*",
+        "ext-pdo_sqlite": "*",
```
samsonasik, line 23: we don't use pdo-sqlite afaik

samsonasik: when install sqlite3 extension, it seems install pdo_sqlite as well, if it need to be here, it may need to be moved under `"require"`

resources/views/base.blade.php
```diff
@@ -7,7 +7,7 @@
         @include('_snippets/layout/head')
         @include('_snippets/styles')
 
-        <link rel="icon" type="image/x-icon" href="favicon.ico" />
+        <link rel="icon" type="image/x-icon" href="{{ asset('favicon.ico') }}" />
```
samsonasik, line 10: I think this can be moved to separate PR, easier to review/merge 👍

.env.dist
```diff
@@ -3,5 +3,6 @@ APP_DEBUG=1
 
 APP_KEY=base64:c9yprqAy56EwJ1pvPeyRkP6xV3MTtsa498Hsffd3/rw=
 APP_URL=http://localhost
+ASSET_URL=/
```
samsonasik, line 6: there are 2 `.env`:

```
.env.dist
.env.prod
```

this may need update `.env.prod` as well for production environment.

PR #3663, opened 2026-04-10
Rename `maxNumberOfProcesses` -> `maxNumberOfProcess`

JHWelch:
In the codebase this argument is `maxNumberOfProcess`. 

https://github.com/rectorphp/rector/blob/8268abaeb93c70c445cef8d4d6cbc76997c39350/src/Config/RectorConfig.php#L84

If you copy and paste this code block it does not run without modification.


samsonasik:
Thank you @JHWelch

